PR #20, opened 2016-03-03
added syntax highlighting for code examples

steveoh:
we need this for

```js
var code = 'examples';
```


steveoh:
this is garbage.

PR #27, opened 2016-03-08
Updates to data language on front page

BGranberg:
Done for the day...ready to merge

### DIFF
--- a/index.html
+++ b/index.html
@@ -67,11 +67,11 @@ layout: default
         </div>
         <div class="grid__col grid__col--1-of-4 text-center">
             <a href=""><h5>Aerial Photos</h5>
-            <img src="holder.js/75x75" height="75" width="75"></a>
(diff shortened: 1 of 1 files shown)


BGranberg:
Ready to merge....mostly work to the front page

BGranberg:
oops...meant to just comment sorry for the unnecessary/unintended close

steveoh:
This looks good. I think we should add the class **grid__col--s-1-of-2** to the pictures on the data section. It will look better on mobile. Will you do that and then I will accept?

steveoh:
I do not like that we are scaling images with html attributes like 

```html
height="75" width="75"
```
this is a bad practice. The image should be 75x75 if that is what size we want them to be. We are wasting bandwidth and slowing down the page load by downsizing a bigger image.

steveoh:
i took care of it.

PR #86, opened 2016-04-12
added projection information

rkelson:
BTW the Code & Description section using dl,dt,dd is not very appealing in case you have a better idea of how to do it.


steveoh:
create an issue for the thing that isn't appealing to you.

PR #91, opened 2016-04-13
added fa-map

rkelson:
1st attempt before duplicating on other pages


steveoh:
for [accessibility purposes](http://fontawesome.io/icon/map/) all the icons should be like this

```html
<i class="fa fa-map" aria-hidden="true"></i>
```

PR #93, opened 2016-04-14
Added hyper link to Blue Stakes

jpechmann:
I added a hyper link to Blue Stakes since James requested it. Maybe we should add hyper links to everything?


steveoh:
@BGranberg do you want to put links everywhere?

PR #104, opened 2016-04-19
Added a better thumbnail for the Economy

jpechmann:
The Economy data category needed a better image.


steveoh:
Are you missing an image here?

PR #128, opened 2016-05-05
Patch 1

rkelson:
fixed code/description


data/elevation-terrain-data/2013-2014-lidar/index.html
```diff
@@ -121,20 +121,13 @@
...
-        <dd>Bare-earth ground</dd>
-        <dt>7</dt>
-        <dd>Noise</dd>
-        <dt>9</dt>
-        <dd>Water</dd>
-        <dt>10</dt>
-        <dd>Ignored Ground (Breakline proximity)</dd>
-        <dt>11</dt>
-        <dd>Withheld</dd>
+      <dt>Code | Description</dt>
```
steveoh: this isn't the semantically correct use of a definition list. We should probably replace this with something more appropriate.

PR #193, opened 2016-08-03
embed agol webmap

ZachBeck:
Pull request description
------------------------


steveoh:
I think there is a sharing issue with this. Let me know when that is resolved.

PR #203, opened 2016-08-17
update

rkelson:
sorry, did not think I would also have this Co update
------------------------


steveoh:
in the future try to author your commit messages in this format "this commit will _____________" omit the this commit will part but try to write it in that format.

rkelson:
so ex. 'update county parcel update date' ?

Rick Kelson
Utah AGRC
801.538.3237
gis.utah.gov

steveoh:
this commit will "update the county parcel info for washington"

PR #205, opened 2016-08-18
Take out link

brigsz:
Pull request description
------------------------

We are trying to improve our update schedule


steveoh:
I would prefer if you deleted the text instead of commenting it out. Are you planning on bringing it back soon? You can always reference this commit to bring it back?

PR #214, opened 2016-08-31
fix relative path

rkelson:
Pull request description
------------------------


steveoh:
thanks. i fixed this already but that is the correct way to do it.

PR #335, opened 2017-01-04
added image for new Metro Townships

mheagin:
added image for new Metro Townships

Pull request description
------------------------


steveoh:
you should be able to do this in the same pull request next time with a commit on the same branch that you opened the pull request with. but great job thanks!

mheagin:
sorry thought I was but than looked and the image addition was not in the same one.  Did I have two forks?

mheagin:
what the fork     should it be two tines then

PR #348, opened 2017-01-13
AGOL Storymap Image Edits

jpechmann:
This pull request replaces 9 images for the UORG storymap with edited images.


steveoh:
These need to be grunt compressed @stdavis

PR #380, opened 2017-02-15
update updates

rkelson:
Pull request description
------------------------


steveoh:
please mention what you are updating. Update updates doesn't help the commit log too much if we were to ever need to use it.

PR #388, opened 2017-02-24
[WIP] Embed UtahMunicipalBoundaries Open Data Page - Landscape

stdavis:


steveoh:
This looks decent. I do not like the header from the open data page. I think we should hide that and place move the link to the open data page to that spot or also remove it. I would also be curious to see how the page looks with the text above the map if that is possible.

steveoh:
I spoke with some open data folks and decided this was not a great/supported idea.

If we want to add previews, we should create a plugin for jekyll and use arcgis online web map id's or feature layer id's to generate a map preview.

brigsz:
Mmmmmm, it is more than just a map preview, more a link to all the other information, like attributes etc. and the other functionality

stdavis:
> I spoke with some open data folks and decided this was not a great/supported idea.

Because design changes on their end breaks our layout? I can see that.

steveoh:
It sounded like we were using an old version of open data and it wasn't supported yet in the new version. I thought it was just for preview so adding a map was the obvious solution

steveoh:
I think we need to wait and see what happens with our open data v2 upgrade. I'm going to close this for now since this is going to be different from v2.

PR #417, opened 2017-03-13
Update to tax data

mheagin:
Updated some email links, The down load section in Entities and Transit special tax areas. Finally update dates for Special Transit

Pull request description
------------------------


steveoh:
this page has display problems that I'm going to need to fix but thank you for updating what you did and bringing it to my attention.

PR #458, opened 2017-05-03
add imagery and italicize free

rkelson:
- [x] additionally Bert would like the logo centered at the top
- [ ] also requested is the Google and NAIP dates under the available services. If Master Po could figure out a good way to do that it would be appreciated. They are:
- [x] NAIP - 2006, 2009, 2011, 2014, 2016
- [x] Google - 2011, 2012, 2013, 2014, 2015, 2016
NAIP

Pull request description
------------------------


steveoh:
merged

PR #487, opened 2017-06-06
add could be outdated info to posts

rkelson:
Pull request description
------------------------


rkelson:
the outdated info should be fixed now

PR #518, opened 2017-07-14
Recreation data links updated to drive

keatonwalker:
Pull request description
------------------------

Updated all the recreation links to drive.


keatonwalker:
urban-trials was not update because the data doe not exist in the SGID and I don't think [the urban-trials html page](https://github.com/agrc/gis.utah.gov/blob/master/data/recreation/urban-trails/index.html) is used anywhere.

PR #527, opened 2017-07-18
clean up

stdavis:
Pull request description
------------------------


steveoh:
You have to be careful with your beautify. It messes up the liquid template links. Check the spacing or we will have broken links.

stdavis:
Ugg. I've just got to quit using beautify on web pages. I think that I've got them fixed.

PR #538, opened 2017-07-19
index links updated to drive

keatonwalker:
Pull request description
------------------------

Index ftp links updated to drive


steveoh:
The sgid-index is generated with code from the google drive stewardship spreadsheet. You'll need to go upstream to fix this file. Let me know after you make those changes so I can generate the page.

PR #555, opened 2017-07-27
clean up and validate html

stdavis:
Pull request description
------------------------

### DIFF
--- a/data/sgid-transportation/index.html
+++ b/data/sgid-transportation/index.html
@@ -6,100 +6,77 @@ author:
   display_name: Bert Granberg
   email: bgranberg@utah.gov
 tags:
-  - airports
-  - airstrips
-  - cartography
-  - class
   - data
-  - ferry
-  - freeways
-  - heavy rail
-  - heliports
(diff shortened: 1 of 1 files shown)


steveoh:
I would also like to rename this folder from `sgid-transportation` to `transportation`. We will need to update the [_redirects](https://github.com/agrc/gis.utah.gov/blob/bfa0e7891d0f43851110f348e7855967682e56a5/_redirects) similar to how it was done with the `society-overview` -> `society` rename

stdavis:
> I would prefer 2 space indent on html but it's not breaker.

I think that it a bit jarring to have html at 2 spaces and other code at 4 spaces but it's not a huge deal to me. I'd like to be consistent however we decide to go. Maybe this is a good case for tabs. Then each user could configure the length to their liking.

stdavis:
> add the overview tag and remove data

Is there a list of acceptable tags anywhere?

stdavis:
OK. I think that I have this cleaned up...

PR #617, opened 2017-09-18
Update embed

brigsz:
Wheeeeeew, I think I know enough to really make a mess now.  I am going to see if I have this someone correct, then fix all the issues

Pull request description
------------------------


steveoh:
Please don't make a pr for each change. Commit to the same banch

steveoh:
That is patch 26

PR #621, opened 2017-09-21
monthly update

ZachBeck:
Pull request description
------------------------


steveoh:
@ZachBeck can you give me the address of a new address point so I can check that the locators were updated?

PR #636, opened 2017-10-05
standardize

stdavis:
I wasn't sure what you wanted to do for these pages. Make changes as appropriate then I can work on the others using this one as a template.


steveoh:
I'd like to merge all of those into one page with all of the years in the grids so you can easily discover what years we offer and download them from a single source.

steveoh:
[This is](https://github.com/agrc/gis.utah.gov/pull/637) what I had in mind. Thoughts?

PR #649, opened 2017-10-30
Online registation is now working

mheagin:
Online registration is working again and we want to add a little more information.

Pull request description
------------------------


steveoh:
thanks!

mheagin:
Sorry need to change one thing "Pay for Renewal" should be just "Register"
oops

PR #673, opened 2017-11-14
Update index.html

brigsz:
deleted an errant html tag

Pull request description
------------------------


stdavis:
That tag is supposed to be there. It was just missing the opening carrot character. It should be `</ul>`.

Issue #611, opened 2017-09-13
Organize contacts in a manageable area


steveoh:
The contacts for stewards and what not change frequently. I think we could use jekyll concepts to organize a list of generic contacts and have jekyll replace all occurrences of them in the website with a simple change. 

We are probably serving emails and phone numbers for outdated or unrelated people.

steveoh:
There are 232 instances of a `mailto:someone` in our website. That seems not ideal. I will try to pull a list.

steveoh:
Here is a list of the folks. @brigsz do any of these stand out as plain wrong at first glance?

```['mailto:jranders@blm.gov', 'mailto:dbuell@utah.gov', 'mailto:katesmith8@gmail.com', 'mailto:sgourley@utah.gov', 'mailto:agrc@utah.gov', 'mailto:gdouglass@utah.gov', 'mailto:david.evans@sgcity.org', 'mailto:leschler@utah.gov', 'mailto:aaronaustin@utah.gov', 'mailto:dbeck@utah.gov', 'mailto:jessicakirby@utah.gov', 'mailto:sfernandez@utah.gov', 'mailto:keatonwalker@utah.gov', 'mailto:jonny.stromberg@gmail.com', 'mailto:michaelfoulger@utah.gov', 'mailto:khansen@jub.com', 'mailto:brooke.conner@fema.dhs.gov', 'mailto:TLA-GIS@utah.gov', 'mailto:zbeck@utah.gov', 'mailto:mpeters@utah.gov', 'mailto:donstaley@utah.gov', 'mailto:rkelson@utah.gov', 'mailto:gbunce@utah.gov', 'mailto:bgranberg@utah.gov', 'mailto:kjohnson@utah.gov', 'mailto:rsempler@utah.gov', 'mailto:tchidester@utah.gov', 'mailto:tschulzke@ufa-slco.org', 'mailto:leeeschler@utah.gov', 'mailto:mheagin@utah.gov', 'mailto:nkenczka@utah.gov', 'mailto:bradhill@utah.gov', 'mailto:mjthomas@utah.gov', 'mailto:lnelms@rideuta.com']```

stdavis:
Might be worth sending an email to that list just to verify that they are good addresses and that the recipients are still valid.

steveoh:
not it!

brigsz:
Yes, there are some that standout as just plain wrong.  How do you want me to edit?

steveoh:
[...]

PR #689, opened 2017-11-29
Contact Management

steveoh:
refs: #611
This creates a `_data/contacts.yml` that has all the contacts in our data pages. 

We can reference contacts in pages with the `_includes/contact.html`
The include has a default text for what it puts before the contact information. You can update that text with the `text` attribute of the include. The `subject` property updates the `mailto` subject link. It is acceptable to use `page.title` unless you would like a more precise name. `hide-punctuation=true` will remove the ending `.` otherwise the include html is all jacked up and if you try to add your own punctuation is get's weird. 

Contacts can be referenced by `site.data.contacts.agrc`.

When using inside of markdown things get a little wonky. You have to create a capture group and then send the output of the capture group through the `strip_newlines` filter because the include html is all weird.

```liquid
{% capture liaison %}{% include contact.html subject=page.title contact=site.data.contacts.luca_liaison hide-punctuation=true text=' ' %}{% endcapture %}
```

otherwise you can do

```liquid
{% include contact.html subject=page.title contact=site.data.contacts.address_grids text='or email'  %}
```

@BGranberg will you please review the `/_data/contacts.yml` to make sure the people are current enough.


_data/contacts.yml
```diff
@@ -0,0 +1,239 @@
...
+  name: Gordon Douglass
+  email: GDouglass@utah.gov
+  phone: 801-538-4810
+  agency: UGS
+water_related_land_use:
+  name: Aaron Austin
+  email: AaronAustin@utah.gov
+  agency: DWR
+voting:
+  name: Mark Thomas
```
BGranberg: Justin Lee

_data/contacts.yml
```diff
@@ -0,0 +1,239 @@
...
+  email: MJThomas@utah.gov
```
BGranberg: justinlee@utah.gov

_data/contacts.yml
```diff
@@ -0,0 +1,239 @@
...
+  agency: DNR-OGM
+oil_fields:
+  name: Brad Hill
+  emal: BradHill@utah.gov
+  agency: DNR-OGM
+air_quality:
+  name: Whitney Oswald
+  email: WOswald@utah.gov
+  agency: DAQ
+brownsfield:
```
BGranberg: brownfields

steveoh:
merged

PR #725, opened 2017-12-21
Application Inventory

steveoh:
Pull request description
------------------------
The start of an application page listing

please review

- [ ] The [Application List](https://deploy-preview-725--gis-utah-gov.netlify.com/developer/applications/)
- [ ] The [crash mapping item](https://deploy-preview-725--gis-utah-gov.netlify.com/developer/applications/vehicle-collision/)


brigsz:
I feel like the Application list is super underwelling, however it is a great start on something that needs to be completed.  I like the Crash Mapping Item, that is nice.  I do not have a good idea yet to enhance the Application list, just thinking, do we group applications by problems they solve or by agency they serve?  The list almost looks to orderly, maybe it just needs styling

steveoh:
Great feedback @brigsz. I think we could remove that list and link to the specific pages from a section like

image

brigsz:
Yes, I forgot about that page, it looks much more ALIVE!

stdavis:
I agree. A list with thumbnails provides an easier way to get an overview of what's going on.

I also think that the image on the crash mapping item page could use a border. It looks a little disjointed with the white background of the image blending in with the rest of the page.

BGranberg:
scott's comment on having more graphics on the index page is valid...I think those table cells could be made tall enough to get a thumbnail in them. And doing that would also allow for short summaries for apps that don't need their own page. Would be great to aim to keep them a uniform height though.

i like the 1 app per line though (the matric view on the front page isn't great if you want to actually digest the info

maybe a column of 'Technologies Employed'  with each cell a list of 'tags' like 'esri, leaflet, giza, mapservAPI'

Issue #608, opened 2017-09-13
Create page for web applications


steveoh:
Give information about website eg: the purpose, contact, links, yada yada and then links to a places where the data can be downloaded.

The work will go in the [gis.utah.gov/developer/applications](http://gis.utah.gov/developer/applications) folder

- [x] Broadband map
- [x] UDAF
- [x] [DEQ](https://gis.utah.gov/developer/applications/deq-environmental/) 
- [x] [locate](https://gis.utah.gov/developer/applications/locate/)
- [x] [raster](https://gis.utah.gov/developer/applications/raster/)
- [x] [WRI](https://gis.utah.gov/developer/applications/watershed-restoration-initiative/)
- [x] [PLSS](https://gis.utah.gov/developer/applications/plss/)
- [x] [crash](https://gis.utah.gov/developer/applications/vehicle-collision/)
- [x] [atlas](https://gis.utah.gov/developer/applications/atlas)
- [x] [parcels](https://gis.utah.gov/developer/applications/parcels/)

steveoh:
[crash download](https://drive.google.com/a/utah.gov/uc?id=18a9jKmFbq2_0zvY9aN5jdMpof5gE0xSG&export=download)

steveoh:
add tracking codes for download data links so we can see if they are getting used.

PR #733, opened 2017-12-27
[WIP] deq application

steveoh:
[preview](https://deploy-preview-733--gis-utah-gov.netlify.com/developer/applications/deq-environmental/) | [html](https://github.com/agrc/gis.utah.gov/blob/enviro-application/developer/applications/deq-environmental/index.html) 

refs #608


stdavis:
Was there a reason why we haven't linked to github repos for any of the projects? I think that it would be a good idea but maybe I missed something.

steveoh:
I think that is a great idea as long as some people are ok with there being a repository in the first place! 😉

PR #764, opened 2018-01-25
update ruby version

steveoh:
Pull request description
------------------------

we don't specify ruby versions. I think the build is using 2.1.2. 2.5 is the latest and in jekyll 4 ruby 2.2 is being dropped. We might be using a lower version than that.

This pr is to see if the site builds ok on the highest version netlify supports.


steveoh:
worked great. waiting to see what version of ruby the docker container uses and hopefully match that.

PR #767, opened 2018-01-26
add 2016 for NAIP

rkelson:
Pull request description
------------------------


steveoh:
do we want to remove the years and assume it is all the naip and hro instead of keeping track of the dates?

rkelson:
no

Rick Kelson
Utah AGRC
801.538.3237
gis.utah.gov

PR #800, opened 2018-03-05
View counter

### LINKED ISSUE #609: Display analytics for page
@BGranberg thought it would be cool to show the analytics for the website or current page somewhere in the page.

Research google analytics to see if there is an api or report we can query on each page load.

really

stdavis:
Pull request description
------------------------
This PR adds a view counter to the bottom of all posts:
image

I'm not even sure if this is still wanted since Bert's not here anymore or if this was the solution that was intended. But it was a great learning experience for me. It's backed by a [google cloud function](https://github.com/agrc/view-counter) that uses Google Datastore. It simply creates a new entity for each unique referrer and increments a count property on each request. No messing with Google Analytics API.

One potential drawback is that all of the pages would start at "1" from the time that this code goes live. So the counts would be incorrect for all but the newest pages. A potential solution would be to write a one-time script that scrapes the data from Google Analytics and pre-populates the datastore with the existing page views.

If we do decide that we want to use it, I wonder if we should stick it in the data pages as well.

This PR also adds IIFEs to the existing JS files. I think that this commit (0b838b8) should be cherry-picked if the PR doesn't end up getting merged.

Closes #609.


steveoh:
This is cool. I like the idea of knowing how often content is viewed. It should be for all pages and posts. Do you want to do the scraping to update the current numbers?

stdavis:
I can give it a try.

stdavis:
OK. I [loaded data](https://github.com/agrc/view-counter/blob/master/scrapeGoogleAnalyticsData.py) from a CSV export of google analytics pageviews into the datastore with the netlify preview domain for this PR just so that we can see how it works. Here's an example: https://deploy-preview-800--gis-utah-gov.netlify.com/sgid-data-lifecycle/

I'll clean out the entire datastore and reimport fresh data (using gis.utah.gov) when this PR is merged.

Next I'll try to figure out how to add this to other pages...

stdavis:
I added this to all pages. Let me know what you think.

steveoh:
Lgtm

stdavis:
I cleaned out the datastore and reimported fresh data (with gis.utah.gov as the domain).

PR #840, opened 2018-04-26
add CSP

steveoh:
https://scotthelme.co.uk/content-security-policy-an-introduction/

### DIFF
--- a/_headers
+++ b/_headers
@@ -4,4 +4,4 @@
   X-Frame-Options: SAMEORIGIN
   X-XSS-Protection: 1; mode=block
   Referrer-Policy: strict-origin-when-cross-origin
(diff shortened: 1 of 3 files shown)


stdavis:
Ya, after reading through the content security blog post, I wondered if this was wise given the amount of stuff that it could break as opposed to the security that we gain. I wasn't sure how it helps our site be more secure since we don't allow any dynamic, user-generated content that I'm aware of.

One issue that has me concerned is that embedded jsbins are broken in the netlify preview but they load fine when I run it on localhost on my machine. There's a good chance of missing issues if they only show up in production but not in development. Is there any way to make development behave more like production?

steveoh:
closing as this was not intended for production

PR #861, opened 2018-06-12
Update 2012-07-11-interactive-map-of-state-fire-restrictions.md

brigsz:
Getting rid of link, that is no longer relevant and also does not work

Pull request description
------------------------


steveoh:
that ksl link is funky also.

PR #876, opened 2018-06-28
delete duplicate links

rkelson:
Pull request description
------------------------


steveoh:
These are duplicated now because you put them in the downloads right?

PR #881, opened 2018-07-06
inhanced information for page

mheagin:
need to push through through these changes and get these pages live, so NV patrons will stop going to TURN to renew.

Pull request description
------------------------


steveoh:
this is great, but we still need the main information on the gis.utah.gov updated.

PR #882, opened 2018-07-09
A few surface edits

atribolet:
Not too much to change here. Just a few surface things! Will leave a couple comments/questions though.

Pull request description
------------------------


mheagin:
morning
Should be "Enoch" city in Iron Co. 
"can affect others" referees to other boundaries; schools, precincts, law enforcement areas....change to ... "can affect other related boundary layers"

PR #885, opened 2018-07-10
Revised GPS network info and link

mheagin:
Change GPS Network heading paragraph and change link to main network page.
Delete anchor point on main page, will link to main network page.

Pull request description
------------------------


steveoh:
That's one way to do it! I'll fix that link

steveoh:
I'm not super excited that we removed that lower panel but ¯\_(ツ)_/¯ oh well. @mheagin @seanfernandez check out the build preview in the main pull request and see if it all looks good please.

mheagin:
Lower panel was too many clicks to get to same info, didn't need a middle man

steveoh:
Sure that is a good point, but the idea with the front page having "middle man" is to teach people what AGRC does "at a glance" by scrolling down the page and not having to browse to other pages to figure that out.

PR #922, opened 2018-08-30
update 2018 lidar blog post

rkelson:
Pull request description
------------------------


steveoh:
I think it's pretty normal to make updates like this a new post with links to the old one and visa versa. You'll get the article in peoples rss/twitter feed/etc this way.

If you already considered this and did not want to do it for some other reasons, I believe I made a front matter item for you that was for updating pages. We should use that instead of editing the original date.

rkelson:
yeah not a lot of changes to need a new post. Guess i'll look for that front matter next week.

PR #936, opened 2018-09-11
Work in Progress: 2018 Utah LiDAR Acquisition Updated

atribolet:
Just a few surface-level changes.

Pull request description
------------------------


atribolet:
@rkelson and @steveoh Just a couple surface changes here, fyi! Shouldn't take too long to update!

PR #940, opened 2018-09-12
SGID SDE Article

atribolet:
Edited! Just needs to be posted! ------------------------


steveoh:
I'm going to close this as it's not current with master and the post is in the wrong location.

PR #957, opened 2018-09-28
Need to update information on page.

mheagin:
Cost is same but now coverage for both networks. 
Need to change the Support Contacts too, not sure where to find this ? ?
Needs to read as follows:
Questions or help call the Support Desk at 801-538-3440 or toll-free 1-800-678-9440.
Comments, compliments, or concerns can be directed to:
Mike Heagin at mheagin@utah.gov
Sean Fernandez at sfernandez@utah.gov

Pull request description
------------------------


steveoh:
closing in favor of #959

PR #978, opened 2018-12-05
update state lidar plan images

rkelson:
Pull request description
------------------------


steveoh:
Dude you are doing great with Github dog!

rkelson:
these image are the same size as the ones they replaced

Issue #981, opened 2018-09-12
SGID.ENERGY.DNROilGasUnits


nathankota:
UDOGM will discontinue updating SGID10.ENERGY.DNROilGasUnits. The source data are actually from the BLM, and provided by 
https://gis.blm.gov/utarcgis/rest/services/Energy/BLM_UT_min_OGUnitAgreements_MapService/MapServer/0

UDOGM think it is still an important layer to maintain in the SGID. Need to rename the layer to indicate its source (perhaps SGID10.ENERGY.BLMOilGasUnits), and have forklift update from the BLM map service.

stdavis:
I transferred this to the gis.utah.gov repo since forklift does not update that layer in SGID. I'll open a PR to update the web page.

steveoh:
The sgid index will need to be updated from the stewardship spreadsheet as well...

stdavis:
OK. Web page and SGID Index is updated. ✅

Should we ask @ZachBeck to remove this dataset from the SGID SDE database?

steveoh:
I believe it is used in pallets? We should check...

ZachBeck:
To me it doesn't make sense to have links to data all over the place and not a central repository.

stdavis:
> I believe it is used in pallets? We should check...

I checked and couldn't find it in any pallets.

steveoh:
Should this stay open?

stdavis:
I'm fine with it closing. Someone just needs to make the decision about what to do with the feature class in SGID. Since we no longer have a great way to update it, I wonder if it should be removed.

nathankota:
Link to BLM data was provided on website, and the layer was removed from the sde, both of which were requested by DOGM. This can be closed.

PR #982, opened 2018-12-06
feat: remove links to SGID data and replace with BLM service

stdavis:
Ref #981


steveoh:
is the sgid index ready to be regenerated?

PR #1007, opened 2019-01-18
Work in progress: Image Services Deprecation Announcement Article

atribolet:
Will leave some quick comments after this too!

Pull request description
------------------------

### DIFF
--- a/_posts/2019-01-17-image-service-deprecation.md
+++ b/_posts/2019-01-17-image-service-deprecation.md
@@ -10,9 +10,9 @@ categories:
 tags: []
 ---
 
-The AGRC brown out of the mapserv image services was very successful. We want to thank you for your participation and apologize for any inconveniences. The few people that were affected and caught by surprise are now informed and on the migration path forward.
+The AGRC brownout of the mapserv image services was very successful. We want to thank you for your participation and apologize for any inconveniences. The few people that were affected and caught by surprise are now informed and are on the migration path forward.
 
-With that behind us, AGRC has chosen the date of **Thursday, February 14** to permanently remove the image services. The deprecation Utah Change Management System reference number is `CHG0019495` and will be discussed at the next two Change Management meetings.
+With the brownout behind us, AGRC has chosen the date of **Thursday, February 14**, to permanently remove the image services. The deprecation Utah Change Management System reference number is `CHG0019495` and will be discussed at the next two change management meetings.
 
 {% capture contact %}{% include contact.html subject=page.title contact=site.data.contacts.change_manager text="If more time is needed for your migration, please let" hide-punctuation=true%}{% endcapture %}
 
--- a/_posts/2019-01-17-image-service-deprecation.md
+++ b/_posts/2019-01-17-image-service-deprecation.md
@@ -18,4 +18,4 @@ With the brownout behind us, AGRC has chosen the date of **Thursday, February 14
 
 We hope this deprecation does not surprise anyone. {{ contact | strip_newlines }} know so we can address them as soon as possible.
 
-As always, please sign up for our newsletter, subscribe to our RSS feed, and follow us on twitter to stay informed about AGRC services and news.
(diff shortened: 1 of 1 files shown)


_posts/2019-01-17-image-service-deprecation.md
```diff
@@ -10,9 +10,9 @@ categories:
 tags: []
 ---
 
-The AGRC brown out of the mapserv image services was very successful. We want to thank you for your participation and apologize for any inconveniences. The few people that were affected and caught by surprise are now informed and on the migration path forward.
+The AGRC brownout of the mapserv image services was very successful. We want to thank you for your participation and apologize for any inconveniences. The few people that were affected and caught by surprise are now informed and are on the migration path forward.
```
atribolet: Fyi, we decided on "brownout" (one word) for the prev announcement article, so I'd recommend consistency with that. 
Also, I'm thinking maybe we should reference that previous article and the brownout a little more specifically to give more context? Like "The January 5, 2019, brownout of image services, as described in XX article..." you know? And then link to it in case people need more background info?

_posts/2019-01-17-image-service-deprecation.md
```diff
@@ -10,9 +10,9 @@ categories:
...
 
-With that behind us, AGRC has chosen the date of **Thursday, February 14** to permanently remove the image services. The deprecation Utah Change Management System reference number is `CHG0019495` and will be discussed at the next two Change Management meetings.
+With the brownout behind us, AGRC has chosen the date of **Thursday, February 14**, to permanently remove the image services. The deprecation Utah Change Management System reference number is `CHG0019495` and will be discussed at the next two change management meetings.
```
atribolet: I'm thinking maybe somewhere too, we should also just briefly mention how people should move forward/prepare for this change? Or should everyone already be ready since they participated in the test brownout? Either way, it might be good to know exactly what might happen when the brownout occurs, you know? Like how should people prepare/what should they expect exactly?

_posts/2019-01-17-image-service-deprecation.md
```diff
@@ -10,7 +10,7 @@ categories:
 tags: []
 ---
 
-The AGRC brownout of the mapserv image services was very successful. We want to thank you for your participation and apologize for any inconveniences. The few people that were affected and caught by surprise are now informed and are on the migration path forward.
+The AGRC brownout of the mapserv image services on mapserv.utah.gov from January 5 through 10, 2019, in preparation for the ArcGIS Server upgrade was very successful. We want to thank you for your participation and apologize for any inconvenience that the brownout may have caused. The few people that were negatively affected and were caught by surprise are now informed about the deprecation of image services and are on the migration path forward.
 
 With the brownout behind us, AGRC has chosen the date of **Thursday, February 14**, to permanently remove the image services. The deprecation Utah Change Management System reference number is `CHG0019495` and will be discussed at the next two change management meetings.
```
atribolet: Also, so we say: "If more time is needed for your migration, please let Matt Peters from AGRC at MPeters@utah.gov know so we can address them as soon as possible."--I think we should revise to "If more time is needed for your migration, please let Matt Peters from AGRC at mpeters@utah.gov know so we can make other arrangements as soon as possible." 

Right now we have a subject/verb agreement with "if more time is needed...so we can address them..." And also, this is more of a general note, could we maybe consider changing the template contact language? I just don't think it reads all that smoothly for some reason. Like, maybe we could switch to something like "...please let Matt Peters (mpeters@utah.gov) from AGRC know..." ? I think it's just the "from" and then "at" that throws me off a little bit.

_posts/2019-01-17-image-service-deprecation.md
steveoh: Yes, I would like to work on the language for the contact include template. It is a little complicated because it is generic and used throughout the entire website. Some contacts don't have certain contact information parts.

Issue #1004, opened 2019-01-15
Add links to AGOL authoritative services


stdavis:
Thanks to the hard work of @MichaelAGRC and others, our AGOL services have become first-class. We should make sure that we have links to all of our authoritative services from their associated data pages and SGID Index on our website. There also may be other AGOL orgs whos services we should link to.

Here's a [list of all AGOL items that have the `sgid` tag and are marked as `authoritative`](https://utah.maps.arcgis.com/home/content.html?view=table&start=1&sortOrder=desc&sortField=modified&tags=sgid&contentstatus=org_authoritative&showFilters=true#content):

- [x] [UtahOilGasWells](http://utah.maps.arcgis.com/home/item.html?id=3bd7bcadd68140b3805e3621d9514294) (prioritized since Don Stayley asked for it)
- [ ] ~~[Municipalities_Modifications](https://utah.maps.arcgis.com/home/item.html?id=572efbac3c8b4799bf729cbeac36dc11)~~ - no match found on website
- [x] [SkiAreaLocations](https://utah.maps.arcgis.com/home/item.html?id=24029712a33f47feacc7b3f2ef75f179)
- [x] [StateFuelSites](https://utah.maps.arcgis.com/home/item.html?id=bbad0291cb754be48ae91dc7d82f6f5c)
- [x] [Utah PLSS Sections GCDB](https://utah.maps.arcgis.com/home/item.html?id=471c7281c52849b2860d36135e751c83)
- [x] [Utah PLSS Townships GCDB](https://utah.maps.arcgis.com/home/item.html?id=5f801b03df2644dd99dcbaa5840bf12d)
- [x] [Utah PLSS Quarter Quarter Sections_GCDB](https://utah.maps.arcgis.com/home/item.html?id=f3454ae613034039b63830c78105bec8)
- [x] [Utah PLSS Quarter Sections_GCDB](https://utah.maps.arcgis.com/home/item.html?id=58b655046e2f4a258dbbf6a6aeaa7b23)
- [x] [Utah Public Safety Answering Point (PSAP 911) Boundaries](https://utah.maps.arcgis.com/home/item.html?id=34f4c65991ce4d4299d3957efa35bf41)
- [x] [Utah Public Safety Answering Point (PSAP_911) Locations](https://utah.maps.arcgis.com/home/item.html?id=6c102c8fc5af413187db0910053d5ed8)
[...]

PR #1013, opened 2019-01-25
feat: implement "Web Services" section in packagedata include

stdavis:
Pull request description
------------------------

This is just an idea of how we could implement AGOL links on the data pages. 

Ref: #1004

image

Example page preview: https://deploy-preview-1013--gis-utah-gov.netlify.com/data/energy/oil-gas/

### DIFF
--- a/_includes/packagedata.html
+++ b/_includes/packagedata.html
@@ -20,6 +20,19 @@
       </ul>
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


steveoh:
This won't work on the pages that don't use the include so we will need to manually add them.

PR #1023, opened 2019-02-07
Final updates for PLSs

mheagin:
the last tow downloads, larger files 1 - the whole fabric, 2, all CadNSDI feature classes for BLM

Pull request description
------------------------


steveoh:
do you want to delete those files from the ftp or is there a reason to keep them there?

mheagin:
don't see a reason to keep them, I have a backup.  That was the last thing we were waiting on, nothing else in there, is there?

steveoh:
Not that I can think of... :fire: it!

PR #1027, opened 2019-02-12
Update contacts.yml to remove David Buell

eneemann:
Updated page to remove Buell and add me for zip_codes and emergencyzones.

Pull request description
------------------------


steveoh:
yay, that was easy!

PR #1061, opened 2019-04-03
remove ', collected within the last 3 years'

rkelson:
add 'Most recent' and 'collected over various years'


steveoh:
Do you think it would be an improvement to put the date ranges of when it was collected, something like between 2015-2017 or something?

rkelson:
if you want. the date of the Google services is 2011 - 2018

Rick Kelson
Utah AGRC
801.538.3237
gis.utah.gov

PR #1095, opened 2019-05-14
change links to new storage

rkelson:


rkelson:
Without setting up a gmail filter to block the [bot] stuff how can I opt
out of getting the lock[bot] clutter?

Rick Kelson
Utah AGRC
801.538.3237
gis.utah.gov

PR #1114, opened 2019-06-04
Add link to original MMNC project description

bertgranberg:


steveoh:
not 100% sure, but i don't think liquid templating works in an include. Also, I think you used the page link syntax when I think what you are referencing is a post. They are a bit different.

bertgranberg:
I probably just need to keep it simple and email you with fixes

-------
Bert Granberg
bert.granberg@gmail.com
SLC, Utah

steveoh:
I'd prefer if you would continue to use github. You can open an issue instead of trying to solve the problem if it is too complicated.

PR #1134, opened 2019-06-25
add 2018 NAIP available

rkelson:
Remove that only mr. sids are available


steveoh:
This good to go now?

rkelson:
I believe so

Rick Kelson
Utah AGRC
801.538.3237
gis.utah.gov

PR #1138, opened 2019-06-25
links to registration not working

mheagin:
Needed to make links say https, not http for the secure link


mheagin:
Thanks, it was a client that was having the issue.  Looks good now

PR #1157, opened 2019-07-09
July updates

ZachBeck:


steveoh:
@ZachBeck will you make a better commit message for things like this. I would suggest: `July address point update: Davis, Salt Lake, Utah, Washington, Weber Counties` 

That would be super helpful!

PR #1164, opened 2019-07-18
add LIR parcel updates

rkelson:
Wasatch
Washington


steveoh:
it'd be dope if you'd put the counties in the commit message instead of the pr description!

PR #1178, opened 2019-07-29
SGID index link

steveoh:
send people to the sgid index from the search page if that's what they wanted.

Does this look good @mheagin?

image


mheagin:
That looks good, points them to the data. 
Can't take them by the hand and lead them. Until they call and either me or Rick have to lead them.

PR #1197, opened 2019-08-16
fix: update QGIS WMTS connection instructions

stdavis:
They were quite out of date.

[QGIS Page Preview](https://deploy-preview-1197--gis-utah-gov.netlify.com/how-to-use-agrc-base-maps-in-qgis/)
[Discover Resources Page Preview](https://deploy-preview-1197--gis-utah-gov.netlify.com/discover/resources/)

I also added a commit of some random files that showed up as changed after I ran the default grunt task. Not sure if you wanted that commit or not.


_posts/2015-01-20-how-to-use-agrc-base-maps-in-qgis.md
```diff
@@ -5,20 +5,19 @@ author:
 tags:
   - base maps
   - open source
-date: 2015-01-20 07:51:04 -0700
+date: 2019-08-16 07:51:04 -0700
```
steveoh: There's an updated front matter tag. Let's use that and keep the original date the same.

PR #1199, opened 2019-08-16
Update AGRC Staff Photo

atribolet:
Not quite sure I did this right, @steveoh ...But I tried to make the photos the same sizes as the ones that were posted, and then the only things I changed were deleting the outdated photo and adding in the new one. (That's where I'm not sure I got it right though? I didn't know how to create a new "Staff" folder for the photos like they were in before, so the file paths looked a little different, and I wasn't sure whether I needed anything else?) Regardless, I tried!


steveoh:
Thanks! you did great, I did a little color correction on the image before merging.

atribolet:
Hooray! Glad it worked out! :) Thanks for fixing it and letting me know!

PR #1227, opened 2019-09-23
Gh image compression action

steveoh:
Great idea! how does this look @stdavis


stdavis:
Would it be worth removing the grunt task also? Or is that still going to be useful?

steveoh:
I think it's ok to keep. I don't think this action compresses all image types.

PR #1232, opened 2019-10-02
Callout for students, info on WMS export problems

jacobdadams:


discover/index.html
```diff
@@ -42,7 +45,7 @@ <h3 class="text-left" id="connect"><i class="fab fa-fw fa-mixcloud"></i> Get Con
   <li>WMTS: <code>https://discover.agrc.utah.gov/login/path/{quad-word}/wmts/1.0.0/WMTSCapabilities.xml</code>
   <li>Tiles: <code>https://discover.agrc.utah.gov/login/path/{quad-word}/tiles/{service}/{z}/{x}/{y}</code>
 </ul>
-<p>For best performance, we recommend using WMTS when possible.</p>
+<p>For best performance, we recommend using WMTS when possible. WMS is slower and has <a href="{{ "/discover/resources/#layers-not-exporting-when-using-wms" | prepend: site.baseurl }}">tehcnical limitations</a> that could prevent the layer from drawing when exporting or printing maps.</p>
```
steveoh:
```suggestion
<p>For best performance, we recommend using WMTS when possible. WMS is slower and has <a href="{{ "/discover/resources/#layers-not-exporting-when-using-wms" | prepend: site.baseurl }}">technical limitations</a> that could prevent the layer from drawing when exporting or printing maps.</p>
```

jacobdadams: Good catch. I'll read up on the new style of links and switch over to those.

steveoh: I updated the readme i think or its in a branch.

PR #1236, opened 2019-10-09
Oct2019 road update

gregbunce:
and update ng911 and network dataset dates


steveoh:
I made a few changes since I think you use a template to update these. If you could update your template to match that would be awesome. I am converting all of our links to use `link` in another branch.

gregbunce:
will do.  thanks!  let me know if I need to do anything - aside from updating my template after your updates.

PR #1257, opened 2019-11-22
Zip Codes page update

eneemann:
Made changes to Zip Codes page for a recent update to create a polygon for Motoqua.


steveoh:
Did this get added to the drive sheet for api geocoding?

eneemann:
I am not familiar with that sheet, so I didn't add it.

gregbunce:
@eneemann I forgot to mention that part.  here's a link to the sheet (https://docs.google.com/spreadsheets/d/1HX47abPQ24LqB8iUHu6HTSmTnbU1YN9WvxO6BDsHCLg/edit#gid=0).  It's essentially a lookup table for zip and address grid.  The api geocoders use it to translate between the two (depending on what's given).  a grid and/or zip can multiple records.  just make sure the new zip has a cross-reference to that grid.  Also, in the weight field, lower number takes precedence.  let me know if you want to chat about any of this.

PR #1269, opened 2019-12-05
add policy page and related content

gregbunce:
this pr will add a new policy section to the website.  this pr will also add the sgid database guidelines and editing policy to that section.  I added a menu item to the top of the page where users can navigate to the policy page, however, I'm unsure if that's the layout we want.  feel free to change that as needed.  if keeping, though, i will need assistance on getting the color-bar under the active page to trigger for "About" section (I moved that menu over one slot).


steveoh:
I wonder if we should put the policies under the about page?

gregbunce:
we certainly could.  i don't know how many "policies" are out there.  if there's a lot, then maybe it warrants it's own menu item, but if not, then yeah, it seems overkill to take up that much real estate on the webpage.  can you think of any other items that could be put out there?

gregbunce:
seems like it could go in the "Quick Links" section of the about page.

steveoh:
I like that idea a lot. I can help but it would be simpler if you made these pr from a branch in this repo instead of a fork. Let me know how you want to proceed

gregbunce:
yeah, i realized i should have done a branch after the fact.

gregbunce:
we're closing this pr and moving it to a new branch called on 'polices' on gis.utah.gov

PR #1270, opened 2019-12-05
add policy page and new sgid policy content

gregbunce:
this pr will add a new page to the website where staff can post related policy. it will be accessed from the 'about' page under the quick links section.


steveoh:
I think we could maybe merge what you did with the about quick links idea and this agrc data disclaimer
image

Adding the cc by 4.0 as well

gregbunce:
sure, I like it.  should i remove the "policies" menu item and add the link to the about page?

steveoh:
I would definitely remove the menu item

gregbunce:
i'm on it (this afternoon).

gregbunce:
done.

gregbunce:
i added the cc by4 tag to the sgid polciy page, but i'm not convinced we need it on this specific page.  i would say it fits better on the data pages.  feel free to keep or remove.  i'm fine either way.

eneemann:
Since this seems to be more of a general, public-facing guideline now, do we need a more detailed policy for the internal editors?  To guide things like:
- Where is the official metadata maintained?
- If internal SDE, in which section do the details go?

gregbunce:
@steveoh I'm done making my initial comments.  it looks good.  i made a few minor suggestions in the code.

PR #1281, opened 2019-12-12
fix ftp links - 2006 Lidar

rkelson:
something went wrong while I was editing so I am not surprised is something is jacked


steveoh:
nice

PR #1295, opened 2019-12-24
update Floodplain downloads

rkelson:
This dataset is an example of downloads that exist on the page that are outside what is available from Opendata. Is the drive folder UtahSGID_Vector eventually completely going away?

### DIFF
--- a/data/water/flood-areas/index.html
+++ b/data/water/flood-areas/index.html
@@ -1,5 +1,4 @@
 ---
-title: Floodplains
 author:
   display_name: JP
   email: agrc@utah.gov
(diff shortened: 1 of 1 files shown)


steveoh:
Which downloads are outside of open data?

rkelson:
there is a link under Updates to where the original data came from.
The Alluvial Fans page is another page with extra download links that currently come from the SGID in Drive

steveoh:
Let’s keep them and drive for now

rkelson:
I figures leave them for now. I'm making a list of pages to go back to when we resolve my question. Greg also has a similar situation to resolve.

steveoh:
If you make that an issue we can all benefit and help fix them later.

PR #1323, opened 2020-01-28
Page edits for SGID updates

eneemann:
Updated descriptions to account for recent SGID data changes in PSAP Boundaries and Locations.


steveoh:
We might need to quote the text because you are using `-` in the yaml.

PR #1332, opened 2020-02-06
Update Public Safety index.html

eneemann:
Updates to the Public Safety page to change the download links from Google Drive to Open Data


eneemann:
Ah, I wasn't sure where that download button link was.  I guess I should've just searched for 'download' or 'google'.  I'll make that change.

PR #1345, opened 2020-02-25
Anchorjs style and removal of duplicate anchor tags

stdavis:
I removed the underline text decoration on the link icons. I also noticed that anchorjs was creating duplicate anchor tags on some pages so I removed those as well.


steveoh:
Did you remove all of the anchor--link items?

stdavis:
Yes, I did this:

image

And then I searched for `anchor--link` and didn't get any results.

PR #1381, opened 2020-04-23
feat: featurey policy

steveoh:
securityheaders.io


steveoh:
image

steveoh:
@stdavis let me know if you notice anything weird with this. I think it should all be ok.

PR #1383, opened 2020-04-28
April updates - round 1

ZachBeck:
Box Elder, Carbon, Davis, Utah, Wasatch, Weber Counties


steveoh:
@ZachBeck do you mind downloading the address points from open data today, and then after the feature layer is updated with these updates, download them again to see if the cache was updated please?

ZachBeck:
No problem

PR #1386, opened 2020-05-04
UTA stops and routes, spring update

ZachBeck:


steveoh:
as you do updates do you want to start migrating to open data for downloads?

PR #1387, opened 2020-05-04
open data downloads for bus stops and routes

ZachBeck:
I might have butchered this one


steveoh:
@jacobdadams will you comment when your script fixes this datasets name please.

PR #1419, opened 2020-06-11
Rkelson patch 1

rkelson:
add links to 3 pages to story map


steveoh:
Your build is fine. i added some plugins not realizing they were active and the build timed out.

PR #1427, opened 2020-06-24
Roadupdate june2020

gregbunce:
draft pr to check the if the collapsible section works in the 911 data page.  note: keep in draft mode until all data is loaded to google drive for download links.


data/911/index.html
```diff
@@ -17,33 +17,37 @@
...
-    - April 12, 2018
-    - March 16, 2018
-    - February 14, 2018
-    - February 6, 2018
-    - January 19, 2018
-    - January 04, 2018
-    - December 18, 2017
+  updated:
+    - June 24, 2020
+    <details>
```
steveoh, line 22: **thought**: i think this should work as it's an html5 feature

data/911/index.html
```diff
@@ -17,33 +17,37 @@
...
+      - August 14, 2018
+      - June 27, 2018
+      - May 30, 2018
+      - April 12, 2018
+      - March 16, 2018
+      - February 14, 2018
+      - February 6, 2018
+      - January 19, 2018
+      - January 04, 2018
+      - December 18, 2017
```
steveoh, line 49: **thought**: i do have a hard time think of reasons to how these old dates provide value

data/911/index.html
gregbunce: i guess it bombed.  maybe b/c it's part of a list item?  ... but, no biggie.  i'm also thinking of removing all the previous dates.  the only reason i leave them is to show we've been updating the ng911 db monthly for X amount of time.

data/911/index.html
gregbunce: see comment above.  but, it looks terrible on the data page.  i might just remove them and put a note in the data description.  i think I'll go that route.

data/911/index.html
steveoh: Ohhh yes, because this is in the front matter. I think you should update the include to add this as a feature! What do you think?

gregbunce: front matter... that's the term i was reaching for.  i bet you're right.  i just axed it in my past commit.  it's kind of overkill.  but, i like the idea of the collapsible list.  i'll probably use it at one point.  thanks for the feedback and troubleshooting.

gregbunce: also, could have bombed b/c i changed the include keyword (for lack of a better term) from "updates" to "updated".  maybe i'll mess with more and use it as a feature on something else when i have more time.

steveoh:
Did you want to migrate this to open data downloads instead of drive while you're making these edits?

gregbunce:
yeah, we may as well add the agol links to roads and trails.  is this what i need to do?
`  endpoints:
    - url: https://utah.maps.arcgis.com/home/item.html?id=XXXXXXXXXXXXXXXX
      label: ArcGIS Online Feature Layer`

gregbunce:
i can't recall.  i don't think i did it last time.

gregbunce:
holy moly, the data is finally updated!  pr is ready.

PR #1436, opened 2020-07-02
Monuments and marker addition

ZachBeck:
This only links to a gdb download. The only way I could think of to link to gdb and shp downloads is to link directly to the open data page.

### DIFF
--- a/data/history/index.html
+++ b/data/history/index.html
@@ -19,9 +19,9 @@ categories: []
       <td>The geographic extent of the Lake Bonneville shoreline.</td>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


stdavis:
You could make two separate links, couldn't you?

ZachBeck:
Would both the links be in the product name area? I'm not sure if I follow.

stdavis:
You could do something like this:
```html
<td>
	Markers and Monuments (<a href="https://opendata.arcgis.com/datasets/63de4118823e4e3eba943ca43caefc72_0.gdb">File GDB</a>, 
	<a href="https://opendata.arcgis.com/datasets/63de4118823e4e3eba943ca43caefc72_0.zip">Shapefile</a>)
</td>
```

PR #1442, opened 2020-07-09
direct trail data links to open data

gregbunce:
and remove google drive links


gregbunce:
it looks good to me - ready for review.

PR #1459, opened 2020-07-29
point landownership data link to sgid opendata

gregbunce:
i was able to work with SITLA and get their AGOL dataset exposed in SGID Open Data.  aka: the backend data on SGID Open Data is SITA's AGOL data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,8 @@ Item:
   hub:
     name: ArcGIS Online name exactly eg Utah Address Points
     item_id: the item id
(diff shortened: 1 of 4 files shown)


gregbunce:
i tested it on org and non-org hub links and it's working as expected. 
Thanks for the assistance and learning experience on this one @steveoh (if i was doing it locally i would have had all those commits from you).

Also, just a side note: i'm off for the rest of day and will return this coming Monday.  

thanks!

PR #1469, opened 2020-08-11
add link to large image

rkelson:
This page was created many years ago and I think some things got lost when we migrated to the GitPages? It most likely will be used in an article the NRCS will be creating about Lidar. I am making some quick changes should I not get time to revamp the page and images by the time the article rolls out. I will know more about the time frame thurs 8/13.


steveoh:
Do you want me to merge this now? This isn't a draft so I'm guessing I should merge but your comment makes me think otherwise. Please advise.

rkelson:
I was just checking the deploy and it looks like the addition I made is
satisfactory. Let's merge.

I dropped the comment so you had an idea why I was messing with the page.
If you get a chance please take a look at the page and see if you can make
it so the Bare-earth DEM image is on the left like the other images. I am
also open to suggestions on revamping the page.

PR #1486, opened 2020-08-28
feat: initial stab at GISAC page

stdavis:
This was a request from @brigsz...


gregbunce:
is the page/site hidden?  i'm just being curious

gregbunce:
nevermind, i'm smarter than that.. i can read git commits.  about/gisac/index.md

stdavis:
@gregbunce yep. It's linked to on https://deploy-preview-1486--gis-utah-gov.netlify.app/about/.

@brigsz Want to take a look? Let me know what needs to be added/changed.

about/index.md
```diff
@@ -84,6 +84,19 @@ AGRC is funded through a combination of State general and restricted funds and c
...
+  </div>
+  <div class="grid__col grid__col--1-of-2" markdown="1">
+###### Utah Geographic Information Systems Advisory Council (GISAC)
+
+- GISAC is composed of individuals from many state agencies.  We are all working together to further the cause of GIS in Utah.  AGRC is trying to create a forum in which all levels of GIS expertise can mingle and learn from each other and learn of each other's challenges. GISAC also sends recommendations to the CIO as potential rules or recommendations that would enhance GIS coordination and cooperation throughout Utah.
+- [Learn more]({% link about/gisac/index.md %})
+{: .dotless }
+  </div>
+  <div class="grid__col grid__col--1-of-2" markdown="1">
+###### National Groups
```
steveoh, line 98: **question**: Why doesn't GISAC fit into one of the local, state, or national groups?

steveoh: I see the change from prior now. Can we update this to a 1-of-3 so it tries to fit on one line?

stdavis: There are actually four groups now:
image

Do you want me to change them to 1-of-4?

steveoh: No I don't think we should have 4 groups. Convince me that GISAC should be is its own group when it appears to be a statewide group. Maybe the group titles should be Local, State, and National to fix the organization. 

The section looks off since the others are only lists of links. Since the description is useful, we should add descriptions for the others or remove the descriptions. 

I'm happy to help grab some descriptions from the group websites or somewhere if you also agree.

steveoh: right, since the main title is GIS Organizations & Affiliates the subheadings are then misleading and incorrect in my opinion. Thanks for your work. Since you aren't interested in sorting this out we'll take it from here.

stdavis: I'm happy to do the work, I just didn't know how to reconcile between what you and @brigsz are asking for. I thought that it would be better for you to hash it out directly rather than try and be the middleman.

steveoh:
@brigsz should we link to the policy pages from the gisac page that we already have in place or is gisac redefining those?

brigsz:
I looked at the page as a preview, it looks good for the moment.  As time goes by I may have more to add to the GISAC page itself, however this gets us started and shows the current task list

brigsz (approved):
I believe this looks good and covers what is needed.  There will be additions in the future

PR #1490, opened 2020-09-04
add road sept2020 road update post

gregbunce:
and update dates for ng911, trials, multimodal network, and statewide network


gregbunce:
it looks good to me.  my only comment is that the trails data on agol open data hasn't caught up yet (i'm guess it will update tonight), but i'd say we merge it b/c all the other data has been updated (2 network datasets; ng911 data; locators; roads downloads).

stdavis (approved):

PR #1508, opened 2020-10-01
Update Data Page to Match SGID Definitions

jacobdadams:
The PR about our new layer policy (#1506 ) has generated some discussion about how we represent what the SGID is on our website. I did some more word work and tried to reword/re-order the main data page to match 4 main parts of the SGID: Index, Open SGID, Open Data, and Raster App. 

I'm hoping this brings some consistency to how we describe the SGID and provides the framework needed to see where the new layer policy fits in.

Edits/rewrites/rotten tomatoes welcome. I could also use some help explaining exactly where the ISO categories come from.

PR #1517, opened 2020-10-09
Notice about impending deprecation of SGID10

jacobdadams:
As I was poking around the site structure for #1508, I came across this page and figured it'd be a good idea to add a warning here for anyone who stumbles on this page. It's probably a limited audience, but still worth a shot?

### DIFF
--- a/sgid/state-only/index.md
+++ b/sgid/state-only/index.md
@@ -11,7 +11,7 @@ layout: page_nocrumb
 categories: []
 ---
 
-**Heads up!** We will be replacing the state-only ArcSDE-based SGID10 database with the [Open SGID]({% link sgid/open-sgid/index.md %}) on **December 31st, 2020**. After this date, the SGID10 database will no longer be accessible. The PostGIS-based Open SGID contains the same datasets and is available to anyone with an internet connection. Please take time now to update your data references.
+**Heads up!** We will be replacing the state-only ArcSDE-based SGID10 database with the [Open SGID]({% link sgid/open-sgid/index.md %}) on **December 31st, 2020**. After this date, the SGID10 database will no longer be accessible. The PostGIS-based Open SGID is available now and contains the same datasets and is available to anyone with an internet connection. Please take time now to update your data references.
 {: .pop .text-center}
 
 The SGID is a collection of vector-format GIS feature classes. Its' content is available for download from the [data pages]({% link data/index.html %}) on this website or via an ArcSDE database connection **for users within the State of Utah network**. Users with ArcGIS desktop clients that are at 10.2.1 or higher can use this resource.
--- a/sgid/state-only/index.md
+++ b/sgid/state-only/index.md
@@ -11,7 +11,7 @@ layout: page_nocrumb
 categories: []
 ---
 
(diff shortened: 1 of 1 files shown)


gregbunce:
i like where you're going with this.  good catch!  

I wonder if we should use stronger wording then "we're planning" since we're on track to remove public access to SGID10 on Dec 31st.

ZachBeck:
My only suggestion would be to refer to the database as SGID10 rather than SGID. Maybe "We will be replacing the SGID10 ArcSDE database accessible within the state network with Open SGID blah, blah..."

jacobdadams:
@ZachBeck and @gregbunce, take another look. I think I was able to do both of your suggestions in the latest version.

ZachBeck:
Looks good to me.

sgid/state-only/index.md
```diff
@@ -10,6 +10,10 @@ title: How to Connect to the SGID via ArcSDE
 layout: page_nocrumb
 categories: []
 ---
+
+**Heads up!** We will be replacing the state-only ArcSDE-based SGID10 database with the [Open SGID]({% link sgid/open-sgid/index.md %}) on **December 31st, 2020**. After this date, the SGID10 database will no longer be accessible. The PostGIS-based Open SGID contains the same datasets and is available now to anyone with an internet connection. Please take time now to update your data references.
```
steveoh, line 14:
```suggestion
**Heads up!** We will be replacing the state-only ArcSDE-based SGID10 database with the [Open SGID]({% link sgid/open-sgid/index.md %}) on **December 31st, 2020**. After this date, the SGID10 database will no longer be accessible. The Open SGID contains the same datasets and is available now to anyone with an internet connection. Please take time now to update your data references.
```

jacobdadams: I think the differences between SDE and PostGIS are important enough to mention, but linking to the blogpost may be more effective than including them here. Let me know what you think about the next commit.

jacobdadams: Also, it might make sense to flesh out sgid/open-sgid with some of the info from the blog post (primarily the differences between PostGIS and SDE and more explanation about the github repo). If anyone sees the Cheshire Cat, let me know.

steveoh:
@jacobdadams is this ready for another review/merge?

jacobdadams:
Yeap, I was just waiting for your review of the latest commit, related to the open suggestion.

sgid/state-only/index.md
```diff
@@ -10,6 +10,10 @@ title: How to Connect to the SGID via ArcSDE
 layout: page_nocrumb
 categories: []
 ---
+
+**Heads up!** As hinted at [earlier this year]({% link _posts/2020-05-04-introducing-open-sgid.md %}), we will be replacing the state-only SGID10 Enterprise Geodatabase with the [Open SGID]({% link sgid/open-sgid/index.md %}) on **December 31st, 2020**. After this date, the SGID10 database will no longer be accessible. The Open SGID contains the same datasets and is available to anyone with an internet connection. Please take time now to update your data references.
```
steveoh, line 14: **nitpick**: 

> The Open SGID contains the same datasets [as the sgid10]

This isn't accurate as we [removed or shelved](https://docs.google.com/spreadsheets/d/1FNCvSQxgQbFG--k3ZapscsTapbrxtSGlKqMcMLCsTYs/edit#gid=0) quite a few layers. I think that should be rephrased

steveoh (commented):
sorry for holding this up. I think once we rephrase the review comment this is ready for approval.

sgid/state-only/index.md
jacobdadams: Makes sense. How does d0758ce25e9b26a9580592bd4bc68ab7dd000044 look to you?

steveoh (approved):

PR #1519, opened 2020-10-09
fix: remove extra character and switch to binary package

stdavis:
Ref: https://pypi.org/project/psycopg2/


eneemann:
Is the binary package a better option?  Also, the "!" was there because it was needed to install packages in Google Colab.  But it's certainly not critical for the blog post.

stdavis:
> Is the binary package a better option?

The binary package doesn't require any prerequisites. Ref: https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary. I think that it would reduce friction for people who are experimenting with this for the first time.

> Also, the "!" was there because it was needed to install packages in Google Colab. But it's certainly not critical for the blog post.

Ah. I never opened the notebook because I was already working on my own project. It wasn't clear to me that the "!" was a Google Colab thing. But maybe it makes sense since you are basically going through the notebook cell-by-cell...

Issue #1521, opened 2020-10-13
Replace drive links with open data


steveoh:
We are taking advantage of the backend provided by open data to provide current downloads in useful formats without having to manage duplicate files. Checkout the [readme](https://github.com/agrc/gis.utah.gov#hub-downloads) for the syntax of hub downloads.

## 911

- [x] [911](https://github.com/agrc/gis.utah.gov/blob/master/data/911/index.html) (not in open data, but moved download to [agrc_public_share](https://drive.google.com/drive/folders/1TVXK10eW6SmblzcUoDq5_VGg4kGDkBvV))

## address-geocoders-locators

- [x] [address-geocoders-locators](https://github.com/agrc/gis.utah.gov/blob/master/data/address-geocoders-locators/index.html) (not in open data, but moved download to [agrc_public_share](https://drive.google.com/drive/folders/1TVXK10eW6SmblzcUoDq5_VGg4kGDkBvV))

## boundaries

- [x] [citycountystate](https://github.com/agrc/gis.utah.gov/blob/master/data/boundaries/citycountystate/index.html)
- [x] [zip-codes](https://github.com/agrc/gis.utah.gov/blob/master/data/boundaries/zip-codes/index.html)
- [x] [wilderness](https://github.com/agrc/gis.utah.gov/blob/master/data/boundaries/wilderness/index.html)
- [x] [school-districts](https://github.com/agrc/gis.utah.gov/blob/master/data/boundaries/school-districts/index.html)
- [x] [conservation-districts](https://github.com/agrc/gis.utah.gov/blob/master/data/boundaries/conservation-districts/index.html)

## cadastre

- [x] [plss](https://github.com/agrc/gis.utah.gov/blob/master/data/cadastre/plss/index.html)

## demographic

- [x] [census](https://github.com/agrc/gis.utah.gov/blob/master/data/demographic/census/index.html)

## economy

- [x] [taxingareas](https://github.com/agrc/gis.utah.gov/blob/master/data/economy/taxingareas/index.html)
- [x] [enterprise-zones](https://github.com/agrc/gis.utah.gov/blob/master/data/economy/enterprise-zones/index.html)

## energy

- [x] [uranium](https://github.com/agrc/gis.utah.gov/blob/master/data/energy/uranium/index.html)
- [x] [renewable-energy](https://github.com/agrc/gis.utah.gov/blob/master/data/energy/renewable-energy/index.html)
[...]

PR #1527, opened 2020-10-15
chore: add hub links

rkelson:
remove endpoints
there are some static links that were moved to the public drive share


rkelson:
just check the links on the deploy and they look like another masterpiece has been created

steveoh (approved):

steveoh:
refs #1521

PR #1544, opened 2020-10-26
chore: switch to v3 api

steveoh:
update to newer download api


steveoh:
spatialRefId is being ignored. always providing 4326. Holding off...

PR #1545, opened 2020-10-27
fix: add space to default text

steveoh:
this should fix the issue.


rkelson (commented):
looks like this will help

PR #1573, opened 2020-11-17
Quaternary faults item id update

ZachBeck:


steveoh (commented):
this is an item shared from another org correct? I don't think this is correct and I think it is appearing to be correct since they share the same name. IF you look at https://github.com/agrc/gis.utah.gov#hub-downloads i think this will need an org property.

steveoh (approved):

Issue #1565, opened 2020-11-10
drive links in non data pages


steveoh:
# posts

- [x] [mapping-spotlight-salt-lake-city-geoportal](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2014-04-25-mapping-spotlight-salt-lake-city-geoportal.md)
- [x] [utah-acquires-high-resolution-aerial-photography-license](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2015-02-02-utah-acquires-high-resolution-aerial-photography-license.md)
- [x] [big-happenings-with-public-land-survey-system-gis](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2015-10-30-big-happenings-with-public-land-survey-system-gis.md)
- [x] [evolving-utah-sgid-web-services-a-proposal](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2015-11-27-evolving-utah-sgid-web-services-a-proposal.md)
- [x] [survey-monument-replacement-and-restoration-county-grant-program](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2016-01-27-survey-monument-replacement-and-restoration-county-grant-program.md)
- [x] [why-i-speak-at-conferences-and-you-should-to](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2016-04-01-why-i-speak-at-conferences-and-you-should-to.md)
- [x] [25th-anniversary-state-geographic-information-database](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2016-05-24-25th-anniversary-state-geographic-information-database.md)
- [x] [data-schema-released-for-expanded-gis-parcel-sharing](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2016-07-07-data-schema-released-for-expanded-gis-parcel-sharing.md)
- [x] [maps-on-the-hill-2017-contest-winners](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2017-02-23-maps-on-the-hill-2017-contest-winners.md)
- [x] [rethinking-geocoders-adding-local-vernacular-into-the-build-process](https://github.com/agrc/gis.utah.gov/blob/master/_posts/2017-06-27-rethinking-geocoders-adding-local-vernacular-into-the-build-process.md)
[...]

PR #1576, opened 2020-11-18
Rkelson drive links round2

rkelson:
I am not sure when I can get back to this so we should probably merge it. I mentioned https://github.com/agrc/gis.utah.gov/issues/1565 in the data meeting yesterday so hopefully people will take ownership of their post.


steveoh:
> I mentioned #1565 in the data meeting yesterday so hopefully people will take ownership of their post.

Wishful thinking! ⭐⭐⭐

steveoh (approved):
looks good. one of the files was in a long term share instead of the blog post data folder but that should be ok right?

Issue #1611, opened 2020-12-30
additions and design of GISAC page


brigsz:
## Issue description

1. I need two links added somewhere to the page.  These are two surveys that GISAC wants to put out.  One will form a directory of GIS users in Utah.  People will register and then it will be open to the public to browse and see others contact info.  https://forms.gle/ea6iPCQ72xwEkRR18

2. The other is for a listing of what your agency has for GIS hardware/software and where it is located. https://forms.gle/tsQvpCbTcUnzESx99 
The last one is just for state of Utah folks (with a login)

3. Now that we have moved beyond just tasks, the web page needs a little organizational face lift.  The Metadata standards will soon link to the ones on gis.utah.gov and now we have the added two surveys, and we may add additional tasks or information based on future meetings etc.
4. Also, somewhere a link to the public notice website for GISAC https://www.utah.gov/pmn/sitemap/publicbody/2172.html
I am working on having someone update the contact info, as it is ancient.

PR #1612, opened 2020-12-30
improve gisac page

steveoh:
refs #1611


steveoh:
Alright @brigsz this is ready for review. I ended up adding the link so as soon as the UPN website is updated it will be live and we don't have to think about it. Go ahead and approve this in review to make it live or leave comments with any content changes or suggestions.

brigsz (approved):

PR #1626, opened 2021-01-12
Libraries hub update

ZachBeck:


data/society/schools-libraries/index.html
```diff
@@ -139,7 +138,7 @@ <h3 id="Libraries">Libraries</h3>
   </div>
   <div class="grid__col grid__col--12-of-12 package-content">
     <img class="inline-text-left outline" src="{% link images/Libraries.png %}" alt="Libraries" loading="lazy" />
-    <p>Society.Libraries provides locations and information for public Libraries and their branches throughout Utah. Includes base location for bookmobile libraries. </p>
```
steveoh, line 142: I suggest we stop using the internal database name here as we are the only ones who can see it. Prefer the open data name?

steveoh (commented):
One change and I think we're good to go

steveoh (approved):

PR #1643, opened 2021-02-02
fix: remove extra padding to nested lists

stdavis:
Before:
image

After:
image


steveoh (approved):

steveoh:
nice and safe. thanks scott

jacobdadams:
The hero we need, but don't deserve 👍

PR #1697, opened 2021-04-08
fix view counter

stdavis:
Update view counter request to match new api in: https://github.com/agrc/view-counter/pull/9


js/source/view-counter.js
```diff
@@ -1,30 +1,27 @@
...
-            return; // don't do anything on request error
-        }
+(async function () {
+  const span = document.getElementById("view-counter");
+  if (!span) {
+    return; // abort if no view-counter element is found
+  }
 
-        var body = JSON.parse(request.responseText);
+  const response = await fetch(
```
steveoh, line 7: **thought**: probably not worth a polyfill

image

stdavis: I agree. If fetch isn't available, the function throws an error and the counter never shows up. The user would never know. Still, that is crazy how much IE there still is out there.

stdavis:
Once this is approved, I'll deploy the cloud function around the same time as I merge this PR.

GruntFile.js
```diff
@@ -60,10 +60,6 @@ module.exports = function (grunt) {
             search: {
                 src: ['node_modules/mustache/mustache.min.js', 'js/source/search.js'],
                 dest: 'js/dist/search.min.js'
-            },
```
steveoh, line 63: **thought**: while i don't really agree i also don't really care :)

steveoh (approved):

GruntFile.js
stdavis: Gaah! Scott's rule number 72, "steve is usually right". I'll work on upgrading Uglify to handle my new code that uses fancy stuff like `const`.

Just for my own enlightenment, what part do you disagree with? Do you really think that we gain any sort of perf from minifying a 20 line file? Or maybe just because it's the right thing to do?

steveoh: For me it's the right thing to do and a trivial task that is not required often.

stdavis: When I tried running uglify after making the changes to the js, it errored out and made me think that it would be a big rabbit hole trying to get it to work. It ended up not being a big deal. This is ready for a final look. I've tested it with the new version of the function and it works.

stdavis:
@steveoh This is ready for another look when you get a chance.

steveoh (approved):

PR #1733, opened 2021-05-12
chore: add culinary water to water category

steveoh:
more than one person looked for culinary water in the water category so let's add it there.


gregbunce (approved):
this looks good.  my only comments is that we'll want to change the "Updates" wording on [this page](https://deploy-preview-1733--gis-utah-gov.netlify.app/data/utilities/retail-culinary-water-suppliers/) to say something like "Updates to this dataset are published periodically."

Issue #1742, opened 2021-06-03
Utah Oil and Gas Well Locations Shapefile Link Broken


jordan-parmer:
## Issue description
The 'Utah Oil and Gas Well Locations' shapefile linking to the following URL is returning a server error:

https://opendata.arcgis.com/datasets/b2d4366fce8d449192b3a0a314bea48e_0.zip"

#### What page is the problem visible
https://gis.utah.gov/data/energy/oil-gas/

#### Expected result
I expected: a shapefile download

#### Actual result
Error page with the text:
`{"errors":[{"title":"Server Error","status":500,"message":"\"_source.item\" is required"}],"meta":{}}`

![Screen Shot 2021-06-03 at 12 41 43 PM](https://user-images.githubusercontent.com/5378348/120688812-1225a800-c469-11eb-8c39-6207dae03370.png)

jacobdadams:
Thanks for the heads up, @jordan-parmer , we're looking into it.

jacobdadams:
In the meantime, you can download them straight from the Open Data page at https://opendata.gis.utah.gov/datasets/utahDNR::utah-oil-and-gas-well-locations/about.

jacobdadams:
@jordan-parmer , we've updated the oil and gas well link and it should be working now from the data page. Give it a try and let us know if its working for you.

jordan-parmer:
Thank you @jacobdadams . It is working now.

PR #1743, opened 2021-06-03
fix: update DNRs AGOL link for wells

jacobdadams:
Should point to the correct service now; for #1742


steveoh (approved):

stdavis:
This id should be stable for now.

PR #1759, opened 2021-06-25
fix: spelling typo

stdavis:
This post is AWESOME, @jacobdadams! Nice tip on the Pro setting. I checked that box in my Pro.


jacobdadams (approved):
Spelling is hard (apparently)

PR #1769, opened 2021-07-01
Revert "fix: replace agrc to ugrc where applicable"

stdavis:
This reverts commit 585c30a495a0e6e77733ef36abac804b8eb4d1e3.

Let's get this pushed to production and then we can re-apply the same commit, fix the yaml formatting issues and then merge again.


jacobdadams (approved):
Takes a deep breath...

PR #1787, opened 2021-07-12
Upgrade Drive Links

stdavis:
This was a bit of a crap-shoot. In some cases it was very hard to determine where the existing file was stored in drive. For most of the links in the blog posts, I moved the files to [`agrc_public_share/blog_post_data`](https://drive.google.com/drive/folders/16wOpp9F00OcFhrRsl5NN15psHEehLG2A). For other pages, I moved some links that seemed obviously old but for others, I took the "if it ain't broke, don't fix it" route.

Also, @rkelson told me not to do anything with indices. Not sure if we want to leave #1565 open until those are sorted out or create a new issue for them.


developer/applications/vehicle-collision/index.html
```diff
@@ -20,10 +20,6 @@ <h3>Tools for visualizing and analyzing vehicle collisions</h3>
         class="fab fa-github"></i><span class="sr-only">github repository link</span> Repository</a>
     <a class="application__repo" href="https://github.com/agrc/crash-db"><i aria-hidden="true"
         class="fab fa-github"></i><span class="sr-only">github repository link</span> Repository</a>
-    <a id="application__download" class="application__download"
```
steveoh, line 23: this is kind of important. why did you remove it? Does it not work?

steveoh (commented):
i'm good with all of this except the removal of the crash data.

developer/applications/vehicle-collision/index.html
stdavis: I was thinking "roadkill" since that's what the link to this page in #1565 showed. I just figured that it was a copy/paste error. I'll put it back.

steveoh: ok, i updated the issue since this isn't roadkill. I don't think we have a spotlight for it in the developer applications section yet

steveoh (approved):

PR #1800, opened 2021-07-26
UTA routes and stops hub name update

ZachBeck:


steveoh (approved):
you didn't need to lowercase the name. only removing the `&` was required.

PR #1807, opened 2021-08-02
stop publishing lidar indices

rkelson:
make page not discoverable


steveoh:
rick wants to give up on this fo rnow.

PR #1810, opened 2021-08-03
docs: CIR band definitions

jacobdadams:
I couldn't find anywhere that we spelled out the band order of the CIR mosiac, so I added it as a parenthetical.


rkelson:
It can't hurt and if it makes you happy it makes me happy

steveoh (approved):

PR #1891, opened 2021-11-30
match congress name with agol

nathankota:
Still getting a 404 on the 2022 congressional districts link to open data. Changed the name to match AGOL.


steveoh:
We can merge this once https://github.com/agrc/porter/issues/162#issuecomment-983131483 is completed.

steveoh:
Actually this is out of date. It should be the other way around.

PR #1910, opened 2021-12-29
post: flattening arrow code

jacobdadams:
A quick walk through of flattening some arrow code. 

I tried making the images resize automatically when the browser size changes with a new CSS class as per https://www.w3schools.com/howto/howto_css_image_responsive.asp, but it didn't seem to be working. Any ideas why, or should I just nix this and merge w/o the dynamic sizing?


steveoh:
> Any ideas why, or should I just nix this and merge w/o the dynamic sizing?

It's probably because of some markdown parsing to html where the `.responsive` class is being applied to a paragraph instead of the image directly. I think it looks good how it is.

jacobdadams:
I've removed the commented code, just needs an approved review to be posted.

steveoh (approved):

PR #1916, opened 2022-01-06
change link & fix text

rkelson:
I can see there are more changes that could be made to this page but there is only so much time in a day


steveoh (approved):

steveoh:
> I can see there are more changes that could be made to this page but there is only so much time in a day

Create some issues describing what could be changed so we don't forget and can make the changes when the days get a little longer.

PR #1920, opened 2022-01-13
SPCS 2022

rkelson:
this is going to take some time to sort out what is need for this


steveoh:
Should `images/Utah_SPCS2022_RegionalZones.png` be  ~2MB or should it be cropped?

PR #1925, opened 2022-01-24
add images for app and upcoming webpage

rkelson:
.


steveoh:
If you could leave a helpful comment besides a period that would help me understand what you are doing. 

Do you want to keep this PR open until you add the app and website stuff or are you using the website to host images for some app and a different website?

rkelson:
We are trying to get somethings together to champion the new spcs stuff. We need to get a few things in order for the committee to review and comment on such as an app and a page.

https://github.com/agrc/gis.utah.gov/pull/1920 will remain a draft until more discussion at the spcs committee level happens and/or I am provided more content for the page. It currently has some images but they might change.

The images in this PR might be used in the draft page but are needed more immediately for app development associated with the spcs page and subject.

steveoh:
Since these images are going to be used in an agol app, let's put them in the agol/apps folder so we can differentiate the use case and make better decisions about when or if we can remove them in the future.

rkelson:
Thanks works for me

steveoh (approved):

PR #1946, opened 2022-02-15
replace Li DAR with Lidar

rkelson:
not sure if this will fix this or if this page auto-generates. If this spelling with the space is coming from the page title I would be happy to change _lidar_ to proper case in the title


steveoh:
You have to update this in the stewardship spreadsheet.

PR #1955, opened 2022-02-22
replace parcel page image

rkelson:
I figured the old image looked like it was generated with ArcView so I created something new. If we agree it is not an improvement I can create a new image.


stdavis:
Do you need to update something else to accommodate the file extension change?

rkelson:
I am not sure how to fix this. After the page would not build, I figured
the issue was the file extension change but not sure. I could delete this
PR and re-add the image as .jpg...tomorrow if need be

steveoh:
If you click the details button here

image

you will be taken to the build log to see helpful messages like this one...

image

steveoh (approved):

PR #1966, opened 2022-03-05
Benefits of testing blogpost

jacobdadams:
I've pulled together some of the benefits I've found from unit testing. Am I missing anything? Anything you would add?

### DIFF
--- a/_posts/2022-02-28-python-shorts-unit-testing.md
+++ b/_posts/2022-02-28-python-shorts-unit-testing.md
@@ -1,0 +1,60 @@
+---
+title: "Python Shorts: The Benefits of Unit Testing"
+author:
+  display_name: Jake Adams
+  email: jdadams@utah.gov
+date: 2021-02-28 11:00:00
+categories:
+  - Featured
+  - Developer
+tags:
+  - python
+---
+
+In the realm of programming, "unit testing" is an automated way of testing small, individual "units" of your code to ensure that when given a specific set of inputs they return a specific set of outputs in the way you expected them to.
+
+If you've never delved into unit testing, you probably won't realize all the ways it can improve your scripts and make you a more efficient programmer in the process.
+
+Here are the six benefits I've discovered since being introduced to unit testing by the other UGRC developers a year and a half ago.
+
+## 1: You Know Your Logic Is Correct
+
+This is the obvious benefit of unit testing. If you write the right tests, you know that each tested part of your code works the way it should. This gives you peace of mind (and job security), especially when whatever your code spits out is being presented to the Mayor.
+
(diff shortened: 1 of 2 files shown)


stdavis (dismissed):
One more thing that might overlap with number 2. If you write the tests first, you are forced to think through how you want to be able to consume your code before you write it.

steveoh:
This is great. Nice job. 

You mention mocking and patching without mentioning what they are. It might be nice to either remove that to keep it simple or quickly explain what that is. 

One thing I really like about testing is how easy it is to try a lot of inputs/cases without having to doctor data or click through a ui. Aka a fast development loop.

jacobdadams:
I've added your suggestions and made a few rewrites; let me know if I captured what you meant.

If everything looks good to you guys, I'll clean up the dates and switch it from draft.

_posts/2022-02-28-python-shorts-unit-testing.md
```diff
@@ -0,0 +1,70 @@
...
+  email: jdadams@utah.gov
+date: 2021-02-28 11:00:00
+categories:
+  - Featured
+  - Developer
+tags:
+  - python
+---
+
+In the realm of programming, "unit testing" is an automated way of testing small, individual "units" of your code to ensure that when given a specific set of inputs they return a specific set of outputs in the way you expected them to.
```
steveoh, line 14: **thought**: the same inputs same outputs thing is called a [pure function](https://en.wikipedia.org/wiki/Pure_function) and also can relate to [idempotence](https://en.wikipedia.org/wiki/Idempotence)

_posts/2022-02-28-python-shorts-unit-testing.md
```diff
@@ -0,0 +1,70 @@
+---
+title: "Python Shorts: The Benefits of Unit Testing"
+author:
+  display_name: Jake Adams
+  email: jdadams@utah.gov
+date: 2021-02-28 11:00:00
```
steveoh, line 6: **chore**: we'll need to freshen up this date and the file name date when you are ready for a merge

steveoh (dismissed):

_posts/2022-02-28-python-shorts-unit-testing.md
jacobdadams: Interesting, thanks for the references. 

If I'm understanding it correctly I don't think I mean to refer to completely pure functions; a test could make sure that an instance method sets a property on its object. My thought was that given a defined state and set of inputs, a function or method produces the expected outputs and side effects. That seems similar to pure functions but broader. Am I understanding that right?

steveoh (approved):

PR #1971, opened 2022-03-10
Open Source Places Blog Post

eneemann:
PR for blog post "Introducing OpenStreetMap Places"

PR #1974, opened 2022-03-16
fix: open source places data page name

eneemann:
I think this is ready for intial review...but we'll still need to update the AGOL ID to point to the newly name layer after forklift works on it over the next couple of nights.


eneemann:
@stdavis please give this a review when you have a chance.  I think it's ready to go and probably worth merging this ASAP and then we can do a simple PR for the new AGOL ID tomorrow.  I'd also request that you help me rebase PR #1971 with the main branch once this is merged. 🙏

data/sgid-index/index.html
```diff
@@ -3279,7 +3279,8 @@ <h6 id='show_filters' class='pointer'><i class="fas fa-search"></i> Show Filters
                 <td data-th="category" class="category">transportation</td>
                 <td data-th="name" class="name"><a href="https://data-uplan.opendata.arcgis.com/search?q=aadt">AADT</a></td>
                 <td data-th="agency" class="agency">UDOT</td>
-                <td data-th="description" class="description">Average Annual Daily Traffic (AADT)  on road sections of the State
+                <td data-th="description" class="description">Average Annual Daily Traffic (AADT)  on road sections of the State
+
```
stdavis, line 3283: Do you know what is causing all of these whitespace changes in this file? Do you have some sort of formatter enabled or something? It would be good if you could remove these changes. It just makes the commit messy. I can help if you need it.

stdavis (changes requested):

data/sgid-index/index.html
eneemann: Hmm...that's weird, I don't think I even touched that line.  It just changed the line related to the open source places.  Should we just revert that file to normal and let PR #1975 (which I thing was auto-created) make that change?

stdavis: That's a good idea. I don't know much about the sgid-index stuff but it seems like this PR shouldn't be touching that file. Let's keep it to #1975.

stdavis (approved):

stdavis:
@eneemann Does this look good to you?

eneemann:
@stdavis Yeah, everything in the preview looks good.  Should I squash and merge?

PR #2012, opened 2022-04-19
feat: housing unit inventory datapages

jacobdadams:
Add a datapage for housing unit inventory with associated entry in the main Planning table.

Marked as draft while finalizing schema, AGOL links.


data/planning/index.html
```diff
@@ -313,6 +313,13 @@
       <td>UGRC</td>
       <td>Locations of Municipal, Private, Public and Resort Golf Courses.</td>
     </tr>
+    <tr>
+      <td>Land Use</td>
+      <td><a href="{% link data/planning/housing-unit-inventory/index.html %}">Housing Unit Inventory</a></td>
+      <td>map data</td>
+      <td>WFRC, MAG, UGRC</td>
+      <td>Characteristics of housing units on a per-parcel or per-development (for PUDs, townhomes, condos, etc) basis derived from LIR parcels and county assessor data.</td>
```
steveoh, line 321: **question**: what is a PUD?

steveoh (commented):
Looks pretty good to me!

data/planning/index.html
jacobdadams: Planned Unit Development- It's a commonly-used way in municipal code to give a developer of a larger area of land more flexibility in lot size/density than the normal, stock zoning would allow. They'll usually have smaller lots with much of the open space grouped together into an HOA-owned common area.

steveoh (approved):

PR #2015, opened 2022-04-20
update OSPlaces index page

eneemann:


steveoh (approved):
the dates are different than the other pages. maybe we can normalize that on the next update.

PR #2028, opened 2022-05-04
GSL 2016

rkelson:
The idea is to have a GSL resources page with the main dataset being the new GSL water level 2016 dataset for the time being


data/water/gsl-water-level-2016/index.html
```diff
@@ -0,0 +1,55 @@
...
+abstract="Various resources related to the Great Salt Lake." %}
+<div class="grid package">
+  <div class="grid__col grid__col--12-of-12">
+    <h3 id="GSL2016">Utah Great Salt Lake Water Level 2016</h3>
+  </div>
+  <div class="grid__col grid__col--12-of-12 package-content">
+    <p>This dataset was derived from the 
+    <a href="https://storage.googleapis.com/state-of-utah-sgid-downloads/lidar/great-salt-lake-2016/GSL_shps.zip">breakline features</a> 
+    associated with the <a href="{% link data/elevation-and-terrain/2016-lidar-gsl/index.md %}">2016 Great Salt Lake LiDAR project</a>. 
+    To create this dataset the <i>Lake Pond</i> features were extracted from the <i>GreatSaltLake_Breaklines</i> shapefile and converted
```
steveoh, line 37: **suggestion**: these may be better as `<code>` blocks instead of `<i>` italics

data/water/gsl-water-level-2016/index.html
```diff
@@ -0,0 +1,55 @@
...
+<div class="grid package">
+  <div class="grid__col grid__col--12-of-12">
+    <h3 id="GSL2016">Utah Great Salt Lake Water Level 2016</h3>
+  </div>
+  <div class="grid__col grid__col--12-of-12 package-content">
+    <p>This dataset was derived from the 
+    <a href="https://storage.googleapis.com/state-of-utah-sgid-downloads/lidar/great-salt-lake-2016/GSL_shps.zip">breakline features</a> 
+    associated with the <a href="{% link data/elevation-and-terrain/2016-lidar-gsl/index.md %}">2016 Great Salt Lake LiDAR project</a>. 
+    To create this dataset the <i>Lake Pond</i> features were extracted from the <i>GreatSaltLake_Breaklines</i> shapefile and converted 
+    into polygons. All polygons outside of the main north and south arms of the lake were removed. The <i>Elevation_m</i> & <i>Elevation_ft</i>
```
steveoh, line 38: **question**: weren't the attributes named elev_m and elev_ft?

steveoh (commented):
looks good so far

data/water/gsl-water-level-2016/index.html
rkelson: i'll take a look at that

data/water/gsl-water-level-2016/index.html
rkelson: actual field name vs. alias and I chose to use the alias. 21st century software defaults to display the alias if it exist I believe. But I could go either way

steveoh: it will only show the alias if consumed in an esri product from the arcgis online feature service or the file geodatabase download right? Every other combination will only have the field name?

rkelson: probably something like that. the page now has both

nathankota:
**questions:** Does it make sense to incorporate the [Lake Bonneville page](https://gis.utah.gov/data/water/historic-lake-bonneville/) into the new GSL page, or should they remain separate? Is [this UGS link](https://geology.utah.gov/popular/general-geology/great-salt-lake/#tab-id-1) a good one to include as a Related Resource?

eneemann:
**question**: should we change the page name to 'gsl-resources' instead of 'gsl-water-level-2016'.  I think that's makes more sense and will match what's on the water page.

data/water/gsl-water-level-2016/index.html
```diff
@@ -0,0 +1,82 @@
...
+</figure>
+{% include abstract.html
+name="Great Salt Lake Resources"
+stewards="UGRC"
+abstract="Various resources related to the Great Salt Lake." %}
+
+<div>
+    <h5>Related Resources</h5>
+    <ul class="dotless">
+      <li><a href="https://wasatchweatherweenies.blogspot.com/2022/04/the-great-salt-finger-lake.html">The Great Salt Finger Lake (2022)</a></li>
```
eneemann, line 39: I think we could do without this one since it's not a data link or state/local government resource. It feels unusual for us to link to a single post on a personal blog.

steveoh: i agree

eneemann:
> not sure how to do this with out merging, copying, duplicating and deleting this. Any ideas?

Looks like you figured out a way 👍

rkelson:
not sure there is anymore to do with this at the moment

steveoh (approved):

PR #2070, opened 2022-06-23
Update DABC to DABS on liquor store page

eneemann:
Update the department name on the liquor store page


stdavis (approved):

steveoh:
Can we make sure that web app gets updated in the stewardship tab also @eneemann

steveoh:
I would add it yah.

PR #2093, opened 2022-07-21
update: OSPlaces HTML

eneemann:
Page update for latest data rebuild.


steveoh (approved):
do you think after 10 updates or so, we should start removing older dates?

eneemann:
Yeah, I was thinking about it this time, but I'll start removing from the bottom next time.

PR #2098, opened 2022-07-27
chore: change quad word

jacobdadams:


steveoh (approved):
should this be bfg'd out of the history?

PR #2186, opened 2022-12-21
update: address data HTML

eneemann:
Added in a minor change for Bluff.

### DIFF
--- a/data/location/address-data/index.html
+++ b/data/location/address-data/index.html
@@ -34,7 +34,8 @@ AddressSystemQuadrants:
   updates:
     - December 2022 - Added several address grids in Beaver county (Elk Meadows, Ponderosa,
       Greenville, Adamsville, Sulphurdale).  Made major updates to grids in Utah, Cache, Tooele,
-      and Box Elder Counties.  Renamed 'NSL' to 'North Salt Lake' and 'East Carbon City' to 'East Carbon'.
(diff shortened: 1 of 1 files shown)


data/location/address-data/index.html
```diff
@@ -34,7 +34,8 @@
   updates:
     - December 2022 - Added several address grids in Beaver county (Elk Meadows, Ponderosa,
       Greenville, Adamsville, Sulphurdale).  Made major updates to grids in Utah, Cache, Tooele,
-      and Box Elder Counties.  Renamed 'NSL' to 'North Salt Lake' and 'East Carbon City' to 'East Carbon'.
+      and Box Elder Counties.  Renamed 'NSL' to 'North Salt Lake' and 'East Carbon City' to 'East Carbon'.
```
steveoh, line 37: **suggestion**: I think you'll want this to be on the same line?

steveoh: actually it all looks kinda the same.

steveoh (approved):

data/location/address-data/index.html
eneemann: Yeah, the long line/multiline stuff always throws me for a loop.  I put it on multiple lines for readability, but it seems like it renders the same anyways.

steveoh: yeah that is html for ya

PR #2187, opened 2022-12-21
update: OpenSourcePlaces HTML

eneemann:
Page update for routine data refresh.


eneemann:
That's weird I'm not sure why it shows that the August 17 line was subtracted and added.

steveoh (approved):

PR #2218, opened 2023-01-31
Adding the link to the update post for 2022

mheagin:
Adding the link to the update post for 2022


steveoh (approved):

steveoh:
It could be nice to add some metadata to the dates. Like which ones are quarterly vs yearly etc.

PR #2254, opened 2023-03-27
update ng911 data date to reflect the latest data

gregbunce:


data/911/index.html
```diff
@@ -17,7 +17,7 @@
     - url: https://drive.google.com/a/utah.gov/uc?id=1Sp6v4FmNV5fhdfEWPTxPGHMbnz-yeY47&export=download
       label: 'UtahNG911GIS: File Geodatabase'
   updates:
-    - January 24, 2023
+    - March 27, 2023
```
steveoh, line 20: **question**: do you want to keep the update history?

steveoh (approved):

data/911/index.html
gregbunce: nah.  we can kill it.  thx for checking.

PR #2286, opened 2023-05-03
uta routes and stops download link fix

ZachBeck:


steveoh:
Sumer didn't want to correct the `_1` issue?

ZachBeck:
for some reason the `_1` doesn't show up in agol https://services.arcgis.com/5QAphMT1g51Tw2X4/ArcGIS/rest/services/UTA_Routes_0419/FeatureServer/128

steveoh:
> for some reason the `_1` doesn't show up in agol [services.arcgis.com/5QAphMT1g51Tw2X4/ArcGIS/rest/services/UTA_Routes_0419/FeatureServer/128](https://services.arcgis.com/5QAphMT1g51Tw2X4/ArcGIS/rest/services/UTA_Routes_0419/FeatureServer/128)

Right, it's an open data issue. It doesn't have anything to do with AGOL really.

steveoh:
Did you and Sumer try to correct the open data issue? It means there is a duplicate service name. If there isn't a duplicate then the service will likely change names again and break our connection. Wouldn't you rather get it corrected now than create more work and issues later?

jacobdadams:
I've been looking at this with Zach. How would we go about finding the duplicate service name? Just asking Summer if there's one in their AGOL org? The REST endpoint doesn't show anything. 

I wonder if hub automatically increments it to _1 when you rename an existing layer as part of the rename, and it will just resolve by itself a little later?

steveoh:
When it happened to us, I think I tried a few things that didn't always fix the issue and sometimes esri support would have to fix it for us.

One thing that has worked when we had actual duplicates, was to delete the agol item and run check for content updates in the content library.

image

I can't remember any of the other toggles to try. But after that it might take a support ticket.

I'll close this to fix the broken links now - thanks for that @ZachBeck.
Thanks for trying to fix the root of the issue @jacobdadams to save ourselves the headache later.

steveoh (approved):

PR #2287, opened 2023-05-03
Basic parcel update

rkelson:
juab
change piute url again


stdavis (approved):
No DNS and HTTP only? 😬 I understand that it's not our issue.

PR #2296, opened 2023-05-11
Newsletter archive update

steveoh:
what do you think about something like this @brigsz


brigsz:
Yep, I think that looks good. Makes it easier to find things

brigsz (approved):
yep, I am happy with that

brigsz:
Hitting lots of buttons, I think they are the correct ones

PR #2302, opened 2023-05-16
feat: specify layer id for downloads, fix uta

jacobdadams:
The download links for gdb/shapefiles appear to use the layer id within the feature service to define what gets downloaded. We've hardcoded in `0` (because we stick to the one layer per service rule), but we have no control over other agencies. This should allow us to specify it when needed, and leave it 0 if not specified.


steveoh (changes requested):
please update the main readme for this new prop

steveoh (approved):

PR #2324, opened 2023-06-14
Add image to 2013 lidar page

rkelson:
I am attempting to add a project area graphic to this html page so it is somewhat like the .md pages
any assistance would be appreciated


steveoh:
the image is quite large but it looks ok

steveoh (approved):

PR #2343, opened 2023-07-21
Update index.html

joshua-reynolds:
This is Josh Reynolds from WFRC Analytics: I changed the paths to the updated projections data & updated the text in Planning page (demographics section)


stdavis (approved):
Thanks!

PR #2359, opened 2023-08-25
feat: organization, sections, trimming

jacobdadams:
I've made my edits on the blogpost, trying to keep the key ideas and concepts. I kept getting lost in the flow of the original, so my edits are more extensive than I thought they'd be.

I organized it a little bit to gather all the problems with the current system in one place and added headings to break it up and guide the reader through the logical flow. I also cut some sections out that didn't relate directly to the deprecation/replacement issue.

I also updated the _deprecated introduction at the beginning per the slack discussion and the GDL diagram.


jacobdadams:
I was hoping I'd get a preview, but I guess that only happens when going to main. We can move it to that if you'd rather.

steveoh:
This has some great changes. I'd like to update some things to my voice but let's get it merged so you can see the preview. Thanks!

jacobdadams:
Have at it 👍

PR #2371, opened 2023-09-22
add shortcut to sgid search index

gregbunce:


steveoh (changes requested):
you duplicated the moth item from 166

gregbunce:
oops, good catch. thanks. fixed.

steveoh (approved):

steveoh:
where are you using these redirects? They might become technical debt with the page structure redesign is the only reason I ask.

gregbunce:
I'm not hard-coding them anywhere.  it's just easier to remember the redirect.  i wanted to put in the nsgic presentation in case anyone wanted to check it out.  so, it would be me telling folks verbally of the URL.

PR #2384, opened 2023-10-13
new fema hq lidar page

rkelson:


rkelson:
I still need to create and add the .png files

steveoh (approved):
I suggest creating a new blog post to create a notification that the 2019 and 2020 products are available. As it is now, no one will know about this accomplishment. It seems worth knocking on some doors about.

rkelson:
yeah maybe one day but not today. The majority of the '19 and '20 projects were completed a couple years ago. This is just the 2020 USGS piece.

steveoh:
Ok, I'll get a tweet out.

PR #2393, opened 2023-10-25
replace hub: with downloads:

rkelson:
this could provide a temporary fix for land ownership downloads


rkelson:
Deploy looks good to me

steveoh (approved):
works for me. who is going to remember to update or replace this when things update and the situation changes? @nathankota

PR #2401, opened 2023-11-20
Update GPS map assets

DenisePeterson:
These updated maps will be used to update the TURN GPS and Nevada GPS network web pages. The maps on those pages are outdated.


steveoh (changes requested):
The title of this pull request is `update outdate map images` but these are all new images. If you are updating images there should be images being removed right?

Are you wanting to replace the maps referenced by these links?

image

If that is the case, please remove the old images and update the links so they point to these new assets all in this same pull request.

gps/turn/index.html
```diff
@@ -31,7 +31,7 @@ <h2 class="text-left">{% include fa_icon.html information=true class="svg-inline
 <p>Eighteen <abbr title="National Geodetic Survey">NGS</abbr> Continually Operating Reference Stations (CORS) were
   selected along the perimeter and interior of the network and were constrained to the published coordinates to conduct
   the final adjustment of the network coordinates.</p>
-<p>With an activated subscription, set your GPS unit's VRS connection address to <code>168.179.231.9</code> and port
+<p>With an activated subscription, set your GPS unit's VRS connection address to <code>165.239.144.5 Port </code> and port
```
steveoh, line 34: looks like you got an extra port in there...
```suggestion
<p>With an activated subscription, set your GPS unit's VRS connection address to <code>165.239.144.5</code> and port
```

gps/turn/index.html
```diff
@@ -97,9 +97,9 @@ <h2 class="text-left">Support Contacts</h2>
   <span class="one-of-two">
     <h2 class="text-left">Related Resources</h2>
     <ul>
-      <li><a href="{% link gps/images/TURN_Network_8x11_2019.jpg %}" rel="noopener" target="_blank">TURN GPS Network Map
+      <li><a href="{% link images/2023TURNGPSNetwork85x11.png %}" rel="noopener" target="_blank">TURN GPS Network Map
           8.5 x 11</a>
-      <li><a href="{% link gps/images/TURN_Network_poster_2019.pdf %}" rel="noopener" target="_blank">TURN GPS Network
+      <li><a href="{% link images/2023TURNGPSNetwork36x44.png %}" rel="noopener" target="_blank">TURN GPS Network
```
steveoh, line 102: after going through another pr and removing old images, let's try to keep these gps based images in the gps/images folder like they were. Do you mind moving them?

DenisePeterson: done.

DenisePeterson:
I'm done working on this for now. Please let me know if you need me to do anything else to move forward with this pull request.

steveoh (changes requested):
this looks great. it's really close. You still need to delete the images you are replacing. 

one other tip would be to swap your commit title and descriptions when you make commits in the future.

As you can see below, the title is some generic item that provides little value but your description is useful. Try replacing the generic title with the useful bits and any extra context can go in the description. Otherwise it's normal to leave the description empty.

image

steveoh (commented):
this looks almost ready to go but I have one last question. 

I'm noticing that there are two unused images. The 20231120*network.png's aren't referenced anywhere. 

What's up with those? Should they be deleted or what's the plan for those?

steveoh (approved):

PR #2402, opened 2023-11-21
basic parcel update

rkelson:
sevier
remove all the 2020 updates


steveoh (approved):
looks good to me. What do you think about only keeping a years worth of updates?

PR #2418, opened 2023-12-20
Update index.html

MichaelAGRC:
This is to update the Watershed  page with the last update date.


data/water/watersheds/index.html
```diff
@@ -12,7 +12,7 @@
     name: Utah Watersheds Area
     item_id: eedaf451685b4cbe875ecf692e1b54ff
   updates:
-    - Unknown
+    - Data updatded from USGS on 11/29/2023
```
steveoh, line 15: **issue**: you've got a spelling issue here and the convention from other pages appears to be Month, Year.

I suggest we remove the `Data updatded from USGS on` part and take that idea and move it south to after the attribute name explanation. Ideally we could link to the page on USGS where we downloaded the data and add a sentence or two of the process to get it into the SGID.

MichaelAGRC (commented):
Added content to enrich the experience.

PR #2423, opened 2023-12-22
fix to allow data downloads

gregbunce:
udot updated the link/service one more time to enable data downloads on this layer.  The previous link/service did not allow data downloads.


gregbunce:
it's good to go

steveoh (approved):

PR #2430, opened 2024-01-02
Remove outdated agrc pic

gregbunce:
2 emps no longer work here, 2 are missing. we need an updated photo.


steveoh (approved):
Thanks Greg. I'd like to use individual head shots so they can be added and removed easier in the new website.

gregbunce:
good idea

PR #2435, opened 2024-01-08
Telecom boundary Porter update

ZachBeck:


steveoh (changes requested):
- [ ] The url should change from `telephone-infrastructure-and-rural-telcom-boundaries` to something more reflective of this new dataset. e.g. `telecom-boundaries`. 
- [ ] the stewardship sheet needs to reflect this url so the sgid index works
- [ ] an entry in `_redirects` should be created for the old and new url so we don't break the internet
- [ ] This utility index page should have the title, description, and link updated
   image
- [ ] The thumbnail is probably misleading
   image

ZachBeck:
What's misleading about the thumbnail, all of those providers are still in the data? It is ugly though...

steveoh:
> What's misleading about the thumbnail, all of those providers are still in the data? It is ugly though...

I assumed it was "probably misleading" since it was created with only the rural telecom boundaries? I'm fine without creating a new one. I wanted to bring it up though. It's not that time consuming to create a new thumbnail either.

steveoh (approved):

PR #2459, opened 2024-01-24
Update index.astro

nathankota:
shorten titles and try to fit all description text in You might also like section


nathankota:
resizing YMAL cards made these changes unnecessary

PR #2460, opened 2024-01-25
Fix Telecom downloads

ZachBeck:
Download links missing

### DIFF
--- a/data/utilities/telecom-boundaries/index.html
+++ b/data/utilities/telecom-boundaries/index.html
@@ -39,6 +39,6 @@ the Public Service Commission." %}
     </ul>
     <p>{% include contact.html subject=page.title contact=site.data.contacts.agrc %}</p>
   </div>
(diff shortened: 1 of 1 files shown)


data/utilities/telecom-boundaries/index.html
```diff
@@ -39,6 +39,6 @@ <h5>Related Resources</h5>
     </ul>
     <p>{% include contact.html subject=page.title contact=site.data.contacts.agrc %}</p>
   </div>
-  {% include packagedata.html name="Telcom"
-  info=page.Telcom %}
+  {% include packagedata.html name="TelecomBoundaries"
+  info=page.TelecomBoundaries %}
```
steveoh, line 43: **issue**: this needs to match line 11

steveoh (changes requested):
good catch. you're only missing an e

steveoh (approved):

PR #2468, opened 2024-02-06
Twitter -> X

stdavis:
I tried to leave unpublished posts alone.


.gitignore
```diff
@@ -1,7 +1,7 @@
 .frontmatter/
 
 # build output
-dist/
+_site/
```
steveoh, line 4: `_site` is the jekyll output. 
you can add `dist` which is the astro output but do not remove it.

`_site` is unnecessary for anyone who has a fresh checkout of the astro website.

stdavis: Oh, I see. I thought that was funny that you wouldn't have that already added. I'll remove that commit.

PR #2470, opened 2024-02-07
delete broken link

rkelson:
found this while pondering the site rebuild. This old page is not like the others and I will need to rework it

I am finding issues on other pages. I will try to clean them up as a group in another PR


steveoh:
thank you!

PR #2472, opened 2024-02-09
Site Search

stdavis:
There's still a few TODOs and a lot of styling that needs to happen.

- [x] general styling to uds spec
- [x] mobile layout
- [x] make work in built app
- [x] figure out how to ignore publish: false blog posts

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -204,6 +204,7 @@
     "owneragency",
     "packagedata",
     "page",
+    "pagefind",
     "palletjack",
     "panguitch",
     "Paragonah",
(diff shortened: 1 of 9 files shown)


steveoh:
We can move the unpublished blog posts into a different area of the website if that would fix an issue

stdavis:
I think that the only outstanding issue that I know of is that if you resize your screen, the responsive menus do not respond. Also, it looks like there is some overlap issues with pages with larger titles:
image

stdavis:
From the accessibility review:
- [ ]  Search button in menu needs a labeled
- [ ]  Close button should be labeled

PR #2484, opened 2024-02-13
Reading estimates for blog posts

stdavis:
The [official recipe](https://docs.astro.build/en/recipes/reading-time/) presents a solution that [other docs suggest is not recommended](https://docs.astro.build/en/guides/content-collections/#modifying-frontmatter-with-remark). I asked about this in the discord and this simpler solution was suggested.

### DIFF
--- a/src/components/page/CardWithDate.astro
+++ b/src/components/page/CardWithDate.astro
@@ -5,11 +5,12 @@ interface Props {
   slug: string;
   date: Date;
   includeYear?: boolean;
+  estimatedReadingTime?: string;
 }
 
 import FormattedDate from '../FormattedDate.astro';
 
-const { title, description, slug, date, includeYear } = Astro.props;
+const { title, description, slug, date, includeYear, estimatedReadingTime } = Astro.props;
 ---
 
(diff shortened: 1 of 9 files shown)


steveoh:
image

Let's remove the reading time from this card. I don't think the width can support the title and the time and they can find out when they nav how long it will take :)

src/components/page/CardWithDate.astro
```diff
@@ -19,7 +20,10 @@ const { title, description, slug, date, includeYear } = Astro.props;
     </div>
     <div class="flex flex-1 pl-6">
       <div class="flex flex-1 flex-col">
-        <p class="line-clamp-1 grow font-semibold text-primary dark:text-zinc-200">{title}</p>
+        <p class="line-clamp-1 grow font-semibold text-primary dark:text-zinc-200">
+          {title}
+          <span class="text-sm">({estimatedReadingTime})</span>
```
steveoh, line 25: this is empty in the deploy preview. I'm suggesting we remove it.

steveoh (approved):

PR #2490, opened 2024-02-14
Blog Index Page Improvements

### LINKED ISSUE #2488: Blog Index Improvements
- [x] full-width layout
- [x] pagination
- [x] filter by category, tag?, author (link to from contact page)
- [x] switch to portrait cards (similar to pillars)

> The trick is that the get static paths function needs to create those pages
> And then you filter the posts by a category or tag and display them
> check out the [category].astro file as an example

stdavis:
Closes #2488


steveoh:
here's an example of someone else's blog post pagination implementation... 

https://github.com/OliverSpeir/astro-starter/blob/main/src/pages/blog/%5B...page%5D.astro

stdavis:
> here's an example of someone else's blog post pagination implementation...

The tags stuff looks cool in that example. But the pagination stuff doesn't look much different from what I already have. Am I missing something?

steveoh:
I didn't look too closely. It doesn't have any redirects.

stdavis:
@steveoh I went a little crazy on this. I need you to look it over and reign me in. Then maybe pair with me on the switch from sticks of gum to baseball cards? Or we could punt on that if there are more pressing concerns.

steveoh:
> @steveoh I went a little crazy on this. 

I really like what you've done so far. It looks super useful

PR #2491, opened 2024-02-15
Docs/discover

jacobdadams:
Here's my first stab at the Discover docs. I'm not married to the way I've broken the topics up, so feel to chime in if you see a better organization.

I took the opportunity to remove a lot of out of date references, like ArcMap.

Also introduces a NumberedListItem component for formatting lists.


rkelson:
@jacobdadams I will take a look at this today. So far it looks great

rkelson:
on the licensed access page we could change the image on the form and use the UGRC logo. Probably also want to remove the _Title_ stuff also.
image

For the Popular Disco Resources the raster app link description could say "Download aerial photography and digital elevation models"

Other than that things look great! Well done

jacobdadams:
Hah, yeah, those forms definitely need some love. 

Thanks for the look-over.

steveoh (approved):
A couple of notes,

1. Use sentence case for section headings.
2. Be careful not to overuse definition lists. Their intended use and screen reader output is to identify a term and give it's definition
3. Make sure to double check for markdown in astro files and convert them to html. e.g. `code` **strong**
4. Use consistent language. Locked down quad-words were also called secure quad-words

PR #2508, opened 2024-02-22
chore: SW user group contact update

nathankota:
Dale Simper replaced by Van Phetsomphou as the user group contact


steveoh (approved):
ideally we migrate these to the contact file in the new website

PR #2521, opened 2024-02-26
feat: place of worship

rkelson:
this is shelved data
there is no record in downloadMetadata.ts for this 
I also did not change the downloads to point straight to opendata 
not sure how these shelved datasets should be addressed


steveoh:
Mark any shelved datasets that you find data pages for in the spreadsheet and skip them

PR #2539, opened 2024-02-29
nhd lakes

ZachBeck:
new nhd lakes page


src/pages/products/sgid/water/nhd-lakes.astro
```diff
@@ -0,0 +1,110 @@
...
+  </Section>
+
+  <Fragment slot="description">
+    <p>
+      UGRC has modified the source NHD data to create additional functionality. Two fields have been added to this
+      feature class (InUtah and IsMajor) and features have been split at the state boundary.
+    </p>
+    <p>
+      <dl>
+        <h3>NHD Codes</h3>
```
steveoh, line 60: **quibble**: headings should be above the `<dl>` and both elements should be outside of the `<p>` tag. Paragraphs should only contain text and links.

PR #2559, opened 2024-03-05
Formatting

stdavis:


.vscode/settings.json
```diff
@@ -376,7 +376,14 @@
     "astro"
   ],
   "[astro]": {
-    "editor.wordWrapColumn": 120
+    "editor.wordWrapColumn": 120,
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
steveoh, line 380:
```suggestion
    "editor.defaultFormatter": "astro-build.astro-vscode"
```

stdavis: Will this create a fight between the two?
image

steveoh: They use each other so I don't think so.

PR #2571, opened 2024-03-06
feat: h3 indices

jacobdadams:
I cheated and added multiple "Use the data" sections, one for each level, instead of duplicating the page five times. I'm not convinced this is the best way to do it, but it seems more concise to me than five nearly-identical pages.


steveoh:
don't forget to update the spreadsheet that you completed this page and the new url

PR #2573, opened 2024-03-06
Jake demographics- hub download refactor

jacobdadams:
Is this a bad idea? 

I'm trying to eliminate the repetitive task for data pages where it makes sense to have multiple items per page (census year pages, probably h3 as well). 

Having multiple download sections does throw off the striping, which needs to be fixed manually. 

But is this just needlessly complicating the components?

### DIFF
--- a/src/components/data/HubDownloads.astro
+++ b/src/components/data/HubDownloads.astro
@@ -1,52 +1,35 @@
 ---
-import type { IHubDownload } from '@models/products/sgid/types';
-
 import Alert from '@components/page/Alert.astro';
-import ExternalLink from '@components/page/ExternalLink.astro';
 
-import Clipboard from '@components/page/Clipboard.astro';
+import HubDownloadsLinksOnly from './HubDownloadsLinksOnly.astro';
 
-const {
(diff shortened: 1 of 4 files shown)


steveoh:
I think it is a bad idea. It's what we wanted to get away from.

jacobdadams:
Is it worse than having 6 separate pages where the only difference is "Census Blocks" instead of "Census Block Groups", etc?

steveoh:
Yes and no. Duplication obviously is an issue, but consistency and familiarity is what we want to provide. Ideally it's an opportunity to explain the uniqueness of each dataset and why you would choose it. It will create special cases for the MATT which also isn't ideal.

jacobdadams:
Hmm... would it be bad to create custom components for each census' description section (2020, 2010, etc) that could then be imported into that section in each separate data page, so we only had one place for the description text?

steveoh:
That's 💯 fine. There's a partials folder where we do this already.

jacobdadams:
closing and nuking branch in favor of `_partials` workflow

PR #2596, opened 2024-03-13
SGID Index Search

stdavis:
First attempt at an interactive SGID index search.

Deploy preview page: https://deploy-preview-2596--gis-utah.netlify.app/products/sgid/sgid-index/


src/components/react/SgidIndexSearchCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+    switch (url.hostname) {
+      case 'dwr-data-utahdnr.hub.arcgis.com':
+        logo = dnrLogo.src;
+        break;
+      case 'data-uplan.opendata.arcgis.com':
+        logo = udotLogo.src;
+        break;
+    }
+  } else {
+    urlString = (import.meta.env.DEV ? 'http://localhost:4321' : astroSite?.href) + urlString;
```
steveoh, line 46: **issue**: this is causing a double slash

image

src/components/react/SgidIndexSearchCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  return (
+    <div className="flex grow flex-col">
+      {record.productPagePath.length > 0 ? (
```
steveoh, line 51: **though**: what do you think about if `record.productPagePath.length > 0`, we then choose the hub, server, or inaction url to use instead (in that order).

image

steveoh (commented):
this is awesome! A couple comments and let's do it!

src/components/react/SgidIndexSearch.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <div className="space-y-6 overflow-x-hidden">
+        {results
+          .filter((result) => {
+            const record = result.item as StewardshipRecord;
+
+            if (categories.length === 0 || record.category === null) return true;
+
+            return categories.includes(record.category);
+          })
+          .slice(0, 15)
```
steveoh, line 89: **polish**: we should add a message saying that there are more results and they were truncated. be more specific yada yada.

PR #2603, opened 2024-03-14
feat: usgs-tile-index page

rkelson:
The Closer Look bullets and striping could use some help

This needs to be added to the 24k hub section but I can't remember how: 
<ExternalLink href="https://utah.maps.arcgis.com/home/item.html?id=baa0bf569643469a9511e1fb753f4a60">USGS 24K Quarter Quads</ExternalLink>.

I don't remember the discussion but at some time Scott pointed out we were using the term 'indices' incorrectly. The page might need edits based on recollection.


steveoh:
yeah drawing outside the lines screws up the striping because it wasn't expected. i'll see what i can do

PR #2604, opened 2024-03-14
Update index.html - Copperton

jidanni:


steveoh (commented):
How does this update relate to UGRC and address grids and benefit anyone reading this information?

steveoh:
I'm choosing to close this as irrelevant. If you disagree please add some context.

jidanni:
I'm worried that the thing that you reviewed said something about Fibonacci numbers.

I didn't write that. I wrote that Copperton now shows two addresses for every house on Google maps.

That means it's in some kind of transition phase.

So the sentence about Salt Lake County having two exceptions, should now say it only has one, as Copperton is probably going to be using the same system as the whole rest of Salt Lake County.

Or maybe it'll actually already is using it. Because I checked the Copperton municipal ordinances and they only referred to Salt Lake County's addressing system.

I'm worried that GitHub didn't even show the word Copperton in any of the things that I changed.

gregbunce:
thanks for the information, jidanni. we'll look into it. i see what you're saying.

jidanni:
Just in case anybody curious why I was researching this, it was for https://wiki.openstreetmap.org/wiki/Utah/Naming_Conventions/Activity , part of https://wiki.openstreetmap.org/wiki/State_of_the_Map_U.S._2024 , which just happens to be taking place in Utah this year. Miss it and you might have to wait another 50 years! Might be a good opportunity to meet local mappers...

steveoh:
Be sure to check out https://gis.utah.gov/the-western-grid/. It is a good read about the topic.

ZachBeck:
Copperton uses the Salt Lake County address grid. The double addresses might be the result of an update that I sent Google last fall. Hopefully the smaller house numbers are a remnant of Google's older, pre-update data and will soon be removed.

steveoh:
@gregbunce and @ZachBeck does that imply that the address grid description for copperton on our site need to be updated?

gregbunce:
i believe so. we would need to remove Copperton from the exceptions list. @ZachBeck do you agree?
https://gis.utah.gov/data/address/address-grids/#salt-lake-county

gregbunce:
thanks, zach! and thanks jidanni for the feedback.  i verified the Address System Quadrants, too and they are correct. 

i believe this issue can be closed.

steveoh:
closed by https://github.com/agrc/gis.utah.gov/commit/3cc55acfe9c214fc8ded9459a96350c31c1b53b6

PR #2614, opened 2024-03-15
Bring external links from category index pages into sgid index

stdavis:
I feel pretty good about the data in the new ["Temp" worksheet](https://docs.google.com/spreadsheets/d/11ASS7LnxgpnD0jN4utzklREgMf1pcvYjcXcIcESHweQ/edit#gid=473227861) other than some manual cleanup.

I think that the next step is to figure out how to get them into the main sgid index page. One idea is to add this data into the migration script and use the `productPage` (I removed the "Path") column to check for duplicates. I'm not sure that this would be great or not.

This also added a `productPageValidation` field to the sgid index and populated it with the new utility function that we wrote.


steveoh (approved):
this looks good to me. What do you want to do about the next steps?

PR #2615, opened 2024-03-15
feat: base map schedule page

### LINKED ISSUE #2530: Query Sheets for Base Map Update Schedule
Replace the google sheets embed with a build-time query for the sheet.

https://gis.utah.gov/developer/base-maps/basemap-maintenance-schedule/

stdavis:
Closes #2530

Deploy preview page: https://deploy-preview-2615--gis-utah.netlify.app/documentation/discover/maintenance-schedule/

### DIFF
--- a/src/utilities/base-maps-dates.ts
+++ b/src/utilities/base-maps-dates.ts
@@ -1,0 +1,49 @@
+import { GoogleAuth, auth } from 'google-auth-library';
+import { GoogleSpreadsheet } from 'google-spreadsheet';
+
+const sheetId = '1XnncmhWrIjntlaMfQnMrlcCTyl9e2i-ztbvqryQYXDc';
+
+const scopes = ['https://www.googleapis.com/auth/spreadsheets', 'https://www.googleapis.com/auth/drive'];
+
+let client;
+if (import.meta.env.PROD) {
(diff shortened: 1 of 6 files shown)


steveoh:
Awesome! Any chance we can pull and add those specific update values for the base map pages I started?

src/utilities/base-maps-dates.ts
```diff
@@ -0,0 +1,49 @@
...
+
+let client;
+if (import.meta.env.PROD) {
+  console.log('using ci credentials');
+  client = auth.fromJSON(JSON.parse(import.meta.env.GOOGLE_PRIVATE_KEY));
+  client.scopes = scopes;
+} else {
+  client = new GoogleAuth({
+    scopes,
+  });
```
steveoh, line 16: **question**: in dev, should we use fake data to avoid rate limiting?

jacobdadams (approved):
I like it. I don't remember my initial thoughts on the concept, but this solution seems really concise.

src/utilities/base-maps-dates.ts
stdavis: Since it's only one call to get the range of cells, I wasn't worried about it.

PR #2620, opened 2024-03-15
Site Map Validation

stdavis:
This PR adds an npm script for site map validation (assuming that the site has already been built).

BTW - I wasn't able find any URLs that are missing from the site map.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "build": "astro build",
     "preview": "astro preview",
     "astro": "astro",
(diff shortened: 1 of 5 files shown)


steveoh (approved):
I'll have to take a closer look at the sheet. Could it be the generated pages? Tags, authors, and categories?

stdavis:
I did a search for all /**/index.html files and compared them with the sitemap. I believe that should catch everything but maybe not.

steveoh:
Your search was in the dist folder after a build?

PR #2632, opened 2024-03-18
Update index.html

ZachBeck:
Updated Salt Lake County address grid description by removing Copperton exception.


steveoh (approved):
Thanks Zach. You'll need to duplicate the change to https://github.com/agrc/gis.utah.gov/blob/main/src/pages/products/sgid/location/address-system-quadrants.astro as well for the new website.

PR #2646, opened 2024-03-19
feat: culinary water suppliers

jacobdadams:
This dataset is hosted in DNR's org but shared through our hub, so I've added a page with an external link to our hub. Not sure if we still need the entry in downloadMetadata, though.

### DIFF
--- a/src/data/downloadMetadata.ts
+++ b/src/data/downloadMetadata.ts
@@ -1012,6 +1012,7 @@ export const dataPages: DownloadMetadata = {
     openSgid: '',
     layerId: 0,
   },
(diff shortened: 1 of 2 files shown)


steveoh:
Is it in the open sgid? Since it's in the agol items table and we have mechanisms to use the hub download with external hubs, I suggest following some of the examples - land ownership, some udot stuff, etc.

jacobdadams:
> Is it in the open sgid? Since it's in the agol items table and we have mechanisms to use the hub download with external hubs, I suggest following some of the examples - land ownership, some udot stuff, etc.

I thought it wasn't in open sgid, but I was wrong. Links are added.

Issue #2659, opened 2024-03-19
Page issues: /products/turn


steveoh:
These two links are missing a location and the images are bogus

<img width="995" alt="image" src="https://github.com/agrc/gis.utah.gov/assets/325813/1ecd602f-fb42-4f1c-be81-cbbebcc9b17a">

DenisePeterson:
I just created a branch with fixes for these issues. If everything looks good, I can close this if you would like.

PR #2673, opened 2024-03-20
fixes for page issues: /products/turn #2659

DenisePeterson:


steveoh (commented):
so fyi most of what you did was unusable. you put the image in the old jekyll folder and referenced it in a way that doesn't work with astro. It's all about the effort!

PR #2695, opened 2024-03-25
Product pillar revisions

jacobdadams:
Review of the products pages (except data pages). I'd especially like @seanfernandez and @DenisePeterson to look over the TURN edits to make sure I'm using `CORS`, etc properly.

### DIFF
--- a/src/pages/_partials/ApiDescription.astro
+++ b/src/pages/_partials/ApiDescription.astro
@@ -18,7 +18,7 @@
 </p>
 <p>
   Searching allows you to explore more than 1,000,000 rows of spatial SGID data. With over 300 layers of real-world data
-  in the SGID, the possibilities can be a little overwhelming. You need a way to sift through all that information and
-  make sense of it, and our search endpoint helps you do that. It opens up the opportunity for serious information
-  gathering to enhance your applications.
+  in the SGID, the possibilities can be a little overwhelming. You need a way to find what you need in all that data,
+  and our search endpoint helps you do that. It opens up the opportunity for serious information gathering to enhance
+  your application and workflows.
 </p>
--- a/src/pages/_partials/ApiDescription.astro
+++ b/src/pages/_partials/ApiDescription.astro
@@ -3,21 +3,22 @@
 ---
 
 <p>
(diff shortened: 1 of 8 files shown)


steveoh (approved):

seanfernandez:
It would sound better if you removed this completely from the products page. "Continuously Operating Reference System (CORS)"

jacobdadams:
What would you change it to, just "reference system"? Because, unless my understanding of GNSS is flawed, GNSS refers to the satellite constellation, ground stations, and receivers, and a reference system like TURN or WAAS are methods to enhance the existing GNSS.

seanfernandez:
In our industry GNSS is like saying GPS except it is referring to a system that uses all available constellation of satellites (USA, Russia, China, Japan, UK) rather than just US satellites. When CORS is used by surveyors it is usually specific to the CORS Network managed by the NGS. 
So it makes sense to use GNSS in place of GPS in this sentence. We originally used GPS because that was the only constellation that was available through our system for a long time, but we have recently upgraded a good portion of the network to provide a full GNSS solution. With that said I don't think we need to go back through everything that mentions GPS and change it to GNSS at this point, but maybe in the future when we are fully integrated.

The Utah Reference Network (TURN) is a high-precision global navigation satellite system (GNSS) network providing real-time corrections and data for post processing. Our network consists of permanently located GNSS receivers installed across Utah and portions of Idaho, Wyoming, and southern Nevada that generate real-time, high-accuracy GNSS positioning. These receivers send their data using the Networked Transport of RTCM via Internet Protocol (NTRIP) to a central system that analyzes the input to adjust and correct the data to provide the most accurate locations, usually down to centimeter level precision. Our network uses the Trimble Pivot VRS Platform.

jacobdadams:
I guess where I'm getting hung up is that by calling TURN a GNSS, to me it
implies that it has it's own constellation, ground stations, receivers,
etc. Maybe this is splitting hairs. I'd really like to call it a GNSS
reference network as a more generic way of saying CORS.

steveoh:
FYI This has  been merged so any new changes will need to be done in a new branch.

seanfernandez:
It is fine for now.

PR #2734, opened 2024-04-22
fix: add link to well paths

stdavis:
Someone emailed today asking where this link went in our new website.


steveoh (approved):
not really a fix but a good addition!

stdavis:
> not really a fix but a good addition!

I saw it as a mistake that it wasn't included. :)

PR #2775, opened 2024-06-04
fix: updates for print proxy v2

stdavis:
Waiting on the production release of Print Proxy v2 for these changes.


steveoh (approved):

src/data/contacts.ts
```diff
@@ -419,11 +419,6 @@ export const contacts: Contacts = {
     email: 'SteveBowman@utah.gov',
     phone: '801-537-3304',
   },
-  print_proxy: {
-    name: 'Scott Davis',
-    agency: 'UGRC',
-    email: 'stdavis@utah.gov',
-  },
```
steveoh, line 426: what was the reasoning to pull out of here?

stdavis: Because I had everything going to ugrc-developers@utah.gov for print proxy.

steveoh: Are you ok with using the contact component and updating the contacts to point to the generic ugrc developers? We have a "rule" not to have mailto links in pages. 

I think we should consolidate our work emails for the group in other places. I can put together that pull request if you're happy with that.

stdavis: I didn't realize that the mailto rule applied to generic accounts but I'm good with that. I'll submit a quick PR to fix this page.

steveoh: thanks. https://github.com/agrc/gis.utah.gov#contact-information

PR #2784, opened 2024-06-24
add interactive map link to post

gregbunce:
folks have asked for a link to app. the embedded map on the page is not great on mobile.


stdavis (approved):

steveoh:
What's up on mobile? Is there some css or otherwise to make it more useful?

gregbunce:
Sounds like navigating is tough, but the main thing is the legend takes over the map.

steveoh:
Should we hide it on small screens?

gregbunce:
Yeah, that would be great if it's possible.

PR #2791, opened 2024-07-08
Update state fuel sites page

rkelson:
this dataset is being removed from the internal database but will persist in agol/OpenData
this is an attempt to replace the standard hub downloads with direct opendata links
add link to app
misc. edits

### DIFF
--- a/src/pages/products/sgid/society/state-fuel-sites.astro
+++ b/src/pages/products/sgid/society/state-fuel-sites.astro
@@ -7,7 +7,6 @@ import { ProductType, SgidCategory } from '@models/types';
 import HubDownloads from '@components/data/HubDownloads.astro';
 import Metadata from '@components/data/Metadata.astro';
 
-import DirectDownloads from '@components/data/DirectDownloads.astro';
 import CardWithPopularLink from '@components/page/CardWithPopularLink.astro';
 import CardWithSmallLink from '@components/page/CardWithSmallLink.astro';
 import ExternalLink from '@components/page/ExternalLink.astro';
--- a/src/pages/products/sgid/society/state-fuel-sites.astro
+++ b/src/pages/products/sgid/society/state-fuel-sites.astro
@@ -48,7 +48,7 @@ const page: IPageMetadata = {
   </Fragment>
 
   <Section title="Use the data" slot="downloads" titlePosition="center">
-    <DirectDownloads title="Explore and Download" links={page.links!} />
+    <HubDownloads {...page.hub} />
   </Section>
 
   <Section title="More resources" slot="more-resources" stripe titlePosition="center">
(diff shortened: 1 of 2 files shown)


src/pages/products/sgid/society/state-fuel-sites.astro
```diff
@@ -26,11 +26,22 @@ export const metadata: IMetadata = {
...
+    },
+    {
+      title: 'File Geodatabase',
+      url: 'https://opendata.arcgis.com/datasets/8d9bc67285fb421fb15d55342148ac1e_0.gdb',
+    },
+    {
+      title: 'Shapefile',
+      url: 'https://opendata.arcgis.com/datasets/8d9bc67285fb421fb15d55342148ac1e_0.zip',
+    },
+  ],
```
steveoh, line 44: i don't understand why this needs to change if it's continuing as an open dataset? What about the internal data influences this change?

rkelson: if it is removed from Internal the Open SGID sample query will be invalid

steveoh: If the [openSgid](https://github.com/agrc/gis.utah.gov/blob/main/src/components/data/HubDownloads.astro#L46) prop from the downloadMetadata is null it won't show that.

rkelson:
this should be all fixed up and ready for review

src/pages/products/sgid/society/state-fuel-sites.astro
```diff
@@ -26,7 +26,8 @@ export const metadata: IMetadata = {
 
 const page: IPageMetadata = {
   ...metadata,
-  updateHistory: [`December 2023`],
+  application: 'https://experience.arcgis.com/experience/9a38f59e6f914367b18fbbde037028e2',
+  updateHistory: [`AS NEEDED`],
```
steveoh, line 30: I would prefer to keep the history instead of changing to as needed.

rkelson: sure but I am the steward and the update history is as needed. Not sure next update I would remember to make the change or that the history was ever maintained or of any significance

steveoh: I understand the refresh cycle is as needed, but that is separate from the update history. We can certainly add the refresh cycle to these pages in the future and that is a good idea. I recommend updating the sgid index to reflect the refresh cycle since it currently says annually.

image

We all need to do a better job of keeping the update log current and accurate. So please try to remember to update the log and the website when you make changes to any dataset. It all plays into the quality metrics of the data and is important to users of the data.

rkelson: duly noted

steveoh (approved):

PR #2806, opened 2024-07-22
change FeatureServiceID to null

rkelson:
for layer deprecated from Internal

hopefully this will fix the index validation error


stdavis:
I don't think that we want to do this. The feature service is not going away, is it?

rkelson:
likely not but I am not sure. All other records in downloadMetadata.ts with `openSgid: ''` have `featureServiceId: null`

steveoh:
The validation error will be fixed with Scott's pr. Sorry for the confusion.

PR #2808, opened 2024-07-22
Update uta datapage links

gregbunce:
this PR updates UGRC's UTA data pages to reflect UTA's new web service links. They recently migrated to arcgis portal and have new links. 

they are also sharing their data with us via distributed collaboration via agol so we no longer need to host their data in Internal. 

this PR is to update the data pages to point folks to their latest and greatest.  the next step will be a Porter issue for the layers we have been hosting (as we no longer need to host this data)

one remaining question is: do we want to serve this new data in Open SGID or are their web services sufficient?


src/data/downloadMetadata.ts
```diff
@@ -1983,17 +1983,17 @@ export const dataPages: DownloadMetadata = {
...
   },
   'Utah UTA Light Rail Stations': {
-    itemId: 'bb8f7bc8871b4a7594bcc037631073e7',
-    name: 'Utah UTA Light Rail Stations',
-    featureServiceId: 'LightRailStations_UTA',
-    openSgid: 'transportation.uta_light_rail_stations',
+    itemId: 'cd624fc1390f439194c8d7b0e3160829',
+    name: 'TRAX Light Rail Stations',
+    featureServiceId: 'TRAX_Stations',
+    openSgid: '',
```
steveoh, line 1989: **chore**: will you update all these empty strings to undefined please

src/data/downloadMetadata.ts
```diff
@@ -1983,17 +1983,17 @@ export const dataPages: DownloadMetadata = {
     layerId: 0,
   },
   'Utah UTA Light Rail Stations': {
```
steveoh, line 1985: **issue**: the key name 'Utah UTA Light Rain Stations' needs to match the name property

steveoh:
These data pages should probably be removed since we aren't hosting the data any longer. Porter requests for each one should be created to deprecate them and then we should use the sgid index/sgid on hub introduction to add them to those places as a distributed collaboration.

gregbunce:
I agree with Steve. I am closing this PR.

PR #2810, opened 2024-07-23
deprecate health care facilities

ZachBeck:
Remove health care facilities in favor of licensed health care facilities


steveoh:
Will you link to the porter issue if this is related to a porter issue please

ZachBeck:
[Porter #256](https://github.com/agrc/porter/issues/256)

steveoh (changes requested):
This seems like a good time in the porter process to remove the page but a couple of things should be done before merging this.

1. The associated downloadMetadata.ts record needs to have the utah health care facilities entry removed.
2. A redirect should be created in the public/_redirects that redirects people to the new product page.

ZachBeck:
I feel like the goal posts are always shifting.

steveoh:
> I feel like the goal posts are always shifting.

I hear you. I wouldn't mind understanding offline what things you feel like are changing often so we can plant the posts in the same place each time. 

The downloadMetadata item is referenced in the page, creating a link between them, and therefore need to be created and removed together. If that doesn't happen we end up with orphan entries and no one wants that.

We don't want to break the web by removing pages. The current page is indexed in search engines etc and to be nice to folks looking for health facilities, we have the opportunity to correct their destination. It's one line in a file. This will always be the case for data replacements but is not necessary for new pages.

I'm hoping that there are maybe 2 or 3 goal posts and we're getting familiar with their location instead of them constantly moving. Again, if that's not the case, i'd like to understand your thoughts offline.

steveoh (approved):

PR #2822, opened 2024-08-13
Update page with link to new CADNSDI data from BLM

DenisePeterson:


src/pages/products/sgid/cadastre/cadnsdi.astro
```diff
@@ -36,8 +37,8 @@ const page: IPageMetadata = {
   ],
   links: [
     {
-      title: 'CadNSDI GIS Version 2.3',
-      url: 'https://drive.google.com/uc?id=13DUod-JnR2W7JLbRM1wjS-bCDrRs93Ij&export=download',
+      title: '2024 CadNSDI GIS Data',
+      url: 'https://drive.usercontent.google.com/download?id=1BiLLawS9uDVpPF-09ISK3M8odQZz2mgb&export=download',
```
steveoh, line 41: This download is wanting utahid authentication. Where is the file located in google?

steveoh (commented):
Once we sort out the download this looks great.

steveoh:
Public data needs to live in the long term share. @rkelson or @gregbunce so you have more details to share?

rkelson:
There is a folder agrc_public_share with blog_post_data, short_term and long_term_share. The short term is kind of like a willcall area and the long term share if for storing things for an indefinite amount of time. I can never find the exact path to agrc_public_share and have a shortcut in Google Drive to it or have a quick access shortcut link to the find it in windows explorer in the G: drive

steveoh:
In private browsing and other browsers i'm not logged into this link works for me. But if i view the deploy preview on mobile safari, I am still prompted for a utah id login whereas on the main website I am not. Can we get a few more testers?

steveoh (approved):

PR #2831, opened 2024-08-16
fix: switch to preferred OGM name

stdavis:
From Megan Crocker at OGM: "Our communications manager has asked us to use this verbiage for the division instead and use OGM for the acronym"


steveoh (approved):
Works for me. We'll need to capture this in our writing guide.

PR #2866, opened 2024-10-09
Ag protection areas update

ZachBeck:
October update


steveoh (approved):
I chatted with Carly and this should be an annual update. I updated the sgid index sheet with that info.

PR #352, opened 2017-01-18
GOED Rec Grants Image

jpechmann:
New cover photo.

Pull request description
------------------------

PR #2869, opened 2024-10-12
Web updates for Porter issue: Remove Utah Metro Townships from SGID #352

DenisePeterson:


steveoh (commented):
A quick scan looks good. Thanks for taking care of this.

I see you updated the blog post for ugrcodes. Are the codes just a spreadsheet or are they a table in the sgid? Is it on your radar to edit that data and any place we've used the codes so everything stays synchronized?

Are we making weekly updates to this data? I assume that change implies no more round up blog posts like we were doing. I would strongly consider with how important this data is to provide more detailed updates than the generic weekly. 

When we remove pages we need to add a redirect so existing links go somewhere valid instead of to a 404 page. I suggest using municipalities as the replacement. This is done by adding a record to the `_redirects` file.

Thanks again for doing this and I'll take a closer look next week.

DenisePeterson:
_I see you updated the blog post for ugrcodes. Are the codes just a spreadsheet or are they a table in the sgid? Is it on your radar to edit that data and any place we've used the codes so everything stays synchronized?_

To my knowledge, the codes are just in a spreadsheet and not in a table in the SGID. It is on my edit and maintain the data so that everything stays synchronized. I will plan to take a look next week to see if I can find any additional pages that need to be updated.

_Are we making weekly updates to this data? I assume that change implies no more round up blog posts like we were doing. I would strongly consider with how important this data is to provide more detailed updates than the generic weekly._

We are making weekly updates to the municipal boundaries data. We will continue doing quarterly blog posts showing updates that were made during that quarter.

_When we remove pages we need to add a redirect so existing links go somewhere valid instead of to a 404 page. I suggest using municipalities as the replacement. This is done by adding a record to the _redirects file._

I have a question about updating the _redirects file so I will reach out to you directly.

steveoh (approved):

PR #2879, opened 2024-10-21
Airport runways pointed at locations

ZachBeck:
Point and line versions of airports were mixed up.


steveoh (changes requested):
Thanks for picking this up. The data page info for airport locations is incomplete. Can you update that as well within this pr please?

https://github.com/agrc/gis.utah.gov/blob/main/src/data/downloadMetadata.ts#L2027-L2033

PR #2888, opened 2024-10-28
update blog post

rkelson:
replace image and delete old


steveoh (changes requested):
There is a, possibly unwritten, convention that all images should be lowercase. This will help avoid these build breaking errors of having mixed case file names. Please delete the new image and add a new one with all lowercase and then update the references to also be lowercase.

steveoh (approved):

PR #2889, opened 2024-10-30
chore: replace gbunce contacts

jacobdadams:
I took over three contacts- census, trails, and luca.

@mhchou, do you want to take on the os places since I think you're running that script now? There may be other stuff in there that you may be the best contact for now as well, that was just what jumped out at me.

### DIFF
--- a/src/data/contacts.ts
+++ b/src/data/contacts.ts
@@ -15,7 +15,7 @@ export const contacts: Contacts = {
   },
   address_points: {
     agency: 'UGRC',
-    email: 'zbeck@utah.gov',
+    email: 'ZBeck@utah.gov',
     name: 'Zach Beck',
     phone: '801-538-3072',
   },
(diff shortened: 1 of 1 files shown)


mhchou:
> @mhchou, do you want to take on the os places since I think you're running that script now? There may be other stuff in there that you may be the best contact for now as well, that was just what jumped out at me.


Yes, I will

steveoh:
Nice work so far, we're chipping away. There's only a couple left and these are my suggestions. Thoughts?

1. @MichaelAGRC for the `address_grids` based on https://github.com/agrc/planning-queue/issues/329.
1. Maybe @ZachBeck and Michael for `roads`
1. For `gisac` I'd suggest @leault 
2. Lastly, I have no idea for `voter_count_tool`. Maybe we can delete this one or use a generic UGRC email?

MichaelAGRC:
Sure.  I'll take the address grids and roads.

leault:
I'll take on gisac.

steveoh (approved):

steveoh (approved):

PR #2923, opened 2024-12-19
Update History

mhchou:
December 2024


src/pages/products/sgid/society/state-agency-dispatch.astro
```diff
@@ -27,7 +27,7 @@ export const metadata: IMetadata = {
 
 const page: IPageMetadata = {
   ...metadata,
-  updateHistory: [`October 2013`],
```
steveoh, line 30: Please keep all the history on product pages. Add back the removed date and we'll get this merged.

mhchou: Added back!

PR #2969, opened 2025-03-06
feat: moth 2024 and 2025

jacobdadams:
Adding the MOTH info for 2024 and 2025. I think I got everything but would appreciate a look-over.


steveoh (approved):
perfect

PR #2994, opened 2025-04-17
republish 2016-09-30-where-and-how-high-resolution-base-maps-and-goog…

rkelson:
…-license.md

updated frontmatter 
links to survey and responses seem to work

this is another good one to republish. It has the some old metrics, the first survey on subject, the first heat map in the survey and survey results


steveoh (approved):
It would be really interesting to make a 2025 version of this with modern metrics!

PR #2999, opened 2025-04-24
Update downloadMetadata.ts

rkelson:
add _undefined_ to openSgid: for Alluvial Fans


steveoh:
@rkelson Is there a specific time that merging this makes sense in the porter process?

rkelson:
@steveoh not sure. With all the moving parts I am not sure that I understand how it all works. I was surprised it was already removed from openSGID

steveoh (approved):

PR #3025, opened 2025-06-10
refactor: migrate podcast episodes to website

steveoh:
- agrc/planning-queue#626

It would be nice to offload these files into a public drive and reference them there. Do either of you have access to the public share?


chriswnek (approved):

chriswnek:
I don't know if I have access to the public share.  If it's a Google Shared drive, it's not in the list of shared drives that I have access to.

PR #3130, opened 2025-11-10
Add links to UGRC Shelf for tax data pages

eneemann:
I created a "tip" to point users to the UGRC Shelf for historical tax datasets from their data page.  Open to other ideas or suggestions, but thought this would be a quick and easy way to point people there.


eneemann:
What did I break?

steveoh:
> What did I break?

You didn't import the new Alert component

steveoh:
And external link. If you click on the failed task, you can see the build log which shows you whats up.

steveoh:
Sorry for the run around. 

We chatted offline and for reasons and prior art to add these links to the more resources section.

I added sorting to the links and to other relevant shelved data. There might be more opportunity with the geoscience datasets but it wasn't very clear to me where they would go.

PR #3132, opened 2025-11-10
fix: law-enforcement-locations update history

eneemann:
Fix the order in the update history to show the most recent update in index position 0 so it's pulled for the top of the page.


steveoh (approved):
sorry i didn't notice this when you formatted it.

PR #3160, opened 2025-12-12
Improve shelved sgid index items

steveoh:
remove no longer valid hub link
remove link to data page
add shelved pill with link to policy

https://deploy-preview-3160--gis-utah-gov.netlify.app/products/sgid/sgid-index/

image

image

### DIFF
--- a/src/components/react/Tooltip.tsx
+++ b/src/components/react/Tooltip.tsx
@@ -17,7 +17,6 @@ const Tooltip = ({ children, content }) => {
         <TooltipPortal>
           <TooltipContent
             className="data-[side=top]:slide-in-from-bottom-2group z-20 max-w-xs rounded-lg border border-slate-800 bg-slate-700 px-3 py-1 text-sm text-white shadow-[inset_0_1px_0_0_theme(colors.gray.600)] drop-shadow-lg will-change-transform data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 dark:border-white/10 dark:bg-slate-600 dark:shadow-none"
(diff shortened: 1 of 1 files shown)


eneemann (commented):
I think this looks really good and I like the shelved "pill" addition.  My only question is does it make sense to include a link to the AGOL item, too, instead of just the service URL?  The AGOL item metadata might provide some additional insight on a specific dataset.  The use case I'm thinking of is a specific tax year or political district year might have additional important context provided in the AGOL item description.

Edit: maybe this a moot comment.  I see the description is on the feature service page, it's just more of a plain text style than the AGOL item page.

eneemann:
> Edit: maybe this a moot comment. I see the description is on the feature service page, it's just more of a plain text style than the AGOL item page.

Looking closer, the feature service description and AGOL item descriptions aren't necessarily the same.  In this case, we have auditor's "shelved" comment in the AGOL item description, but not the feature service description.
image

stdavis:
Thinking more about this...

I'm guessing that many of the users of our site have no idea what "shelved" means. I wonder if there should be a more apparent warning that the data may not be the most current version available. Maybe something like this:

"Shelved: Please note that this data may not be the most current version available"

src/components/react/Tooltip.tsx
```diff
@@ -0,0 +1,34 @@
...
+          </TooltipContent>
+        </TooltipPortal>
+      </Root>
+    </TooltipProvider>
+  );
+};
+
+export default Tooltip;
+
+// data-[state=delayed-open]:data-[side=bottom]:animate-slideUpAndFade data-[state=delayed-open]:data-[side=left]:animate-slideRightAndFade data-[state=delayed-open]:data-[side=right]:animate-slideLeftAndFade data-[state=delayed-open]:data-[side=top]:animate-slideDownAndFade
```
stdavis:
```suggestion
            sideOffset={5}
          >
            {content}
            <TooltipArrow className="fill-bg-slate-700 dark:fill-bg-slate-600" />
          </TooltipContent>
        </TooltipPortal>
      </Root>
    </TooltipProvider>
  );
};

export default Tooltip;

```

stdavis (approved):

eneemann (dismissed):
I think this looks great!  I noticed the same minor visual issue that Scott noted, for the first item in the SGID Index list, but it looks like that's been resolved.  Thanks for putting this in so quickly!

PR #3165, opened 2025-12-16
Feat: new high risk wui data page

eneemann:
Creating a new data page for FFSL's upcoming High Risk Wildland Urban Interface data layer.


stdavis (approved):
I'm assuming that we are OK publishing the OpenSGID table name a few days before it will actually show up in the database? I don't see a huge problem with it.

stdavis (approved):

PR #3315, opened 2026-04-22
feat: add new accessibility page

stdavis:
Ref https://github.com/agrc/planning-queue/issues/960

I originally created a new page for this but it was very small and empty. Then I had the idea of adding it as a new section to our about page and linking to the section. I thought that this worked better and gave it more visibility. I still have the branch that adds it as a new page so if we want to do that it will be easy to switch.


chriswnek (approved):
Looks good, I like the way it looks within the About page.

jacobdadams (approved):

PR #3339, opened 2026-05-11
basic parcel update

rkelson:
salt lake
uintah
utah
washington

refs agrc/planning-queue#905


stdavis:
Nice catch with the comma!

stdavis (approved):

PR #3354, opened 2026-06-02
Fix: Add Alex, remove Jake

acneville:
Replacing monoscale image and `about` object in staff list.


stdavis (approved):
The 1st of many more contributions!

PR #3409, opened 2026-08-20
Update gop-parcels update history

rkelson:
and changed ' to `


stdavis (approved):
`'` or `` ` `` will work for these types of strings. `` ` `` is only required if you want to use a variable within the string like this:

```js
`This data was first made public by UGRC ${date}`
```

